PR #57, opened 2021-02-17
Schedule usage and modification

KasiaKoz:
This is an absolute chonker 😭  don't feel like you need to review the whole thing, I'm leaving comments throughout to point to the more interesting bits

In general:
- changes the underlying structure of the Schedule object a bit. Makes routes and services live on the graph as dictionaries, which makes it easier to find/retrieve them and apply changes quickly (i.e. not iterating through each route/service/stop object one by one and applying changes to them)
- adds methods to summarise, extract data and apply changes to Routes, Services and Stops within a Schedule
- adds convenience methods for modal and spatial extractions (these are brand new and added to Network as a bonus too)
- adds methods to add and remove Services, Routes and Stops


genet/schedule_elements.py
```diff
@@ -26,25 +35,58 @@ class ScheduleElement:
...
+        if service_id in self._graph.graph['services']:
+            return Service(_graph=self._graph, **self._graph.graph['services'][service_id])
+        else:
+            raise ServiceIndexError(f'Service with index {service_id} not found')
+
+    def _get_route_from_graph(self, route_id):
+        if route_id in self._graph.graph['routes']:
+            return Route(_graph=self._graph, **self._graph.graph['routes'][route_id])
+        else:
+            raise RouteIndexError(f'Route with index {route_id} not found')
```
KasiaKoz, line 60: Notice that now you get a Route or Service object from the Schedule when you request it, they all sleep on the graph until called --- just like the Stops, which are just the nodes in the graph.

genet/schedule_elements.py
```diff
@@ -26,25 +35,58 @@ class ScheduleElement:
...
+    def _route_ids_in_graph(self, route_ids: List[str]):
+        return set(route_ids).issubset(set(self._graph.graph['routes'].keys()))
 
-    def _update_graph(self, new_graph):
-        self._graph = new_graph
+    def _service_ids_in_graph(self, service_ids: List[str]):
+        return set(service_ids).issubset(set(self._graph.graph['services'].keys()))
+
+    def change_log(self):
+        return self._graph.graph['change_log']
```
KasiaKoz, line 72: added the changelog to the graph too, so it would be shared across/accessible to all objects within the Schedule

genet/schedule_elements.py
```diff
@@ -267,7 +323,36 @@ def __init__(self, route_short_name: str, mode: str, stops: List[Stop], trips: D
...
+                verify_graph_schema(_graph)
+                self._graph = _graph
+            else:
+                raise RouteInitialisationError('When passing `ordered_stops` you are expected to pass `_graph` too. '
+                                               'You may prefer to pass a list of Stop objects to `stops` instead')
+            # check all stops in _graph
+            if not self._stop_ids_in_graph(ordered_stops):
+                raise RouteInitialisationError('Some stop IDs passed in `ordered_stops` are missing from the _graph '
+                                               'object passed')
+            self.ordered_stops = ordered_stops
```
KasiaKoz, line 345: this is all to do with initiating a route from a shared schedule graph

KasiaKoz: but notice they are hidden in kwargs because it's not something a user should concern themselves with

genet/schedule_elements.py
```diff
@@ -287,22 +372,74 @@ def __str__(self):
 
     def _build_graph(self, stops: List[Stop]):
         route_graph = nx.DiGraph(name='Route graph')
-        route_nodes = [(stop.id, stop.__dict__) for stop in stops]
+        try:
+            route_nodes = [(stop.id, stop.__dict__) for stop in stops]
+            stop_edges = [(from_stop.id, to_stop.id) for from_stop, to_stop in zip(stops[:-1], stops[1:])]
+        except AttributeError:
+            route_nodes = [(stop, {}) for stop in stops]
+            stop_edges = [(from_stop, to_stop) for from_stop, to_stop in zip(stops[:-1], stops[1:])]
```
KasiaKoz, line 380: looks hacky, enables building a route graph only for the purpose of adding it to a schedule graph that already has stops. Don't worry, if they give a stop id that doesnt exist in the Schedule they will get a warning that there isnt any data for that stop when they try to add it (there is no other way to know before you try to add it, as you're only initiating this route her, it has no connection to the Schedule yet)

genet/schedule_elements.py
```diff
@@ -287,22 +372,74 @@ def __str__(self):
...
         route_graph.add_nodes_from(route_nodes, routes=[self.id])
-        stop_edges = [(from_stop.id, to_stop.id) for from_stop, to_stop in zip(stops[:-1], stops[1:])]
         route_graph.add_edges_from(stop_edges, routes=[self.id], modes=[self.mode])
+        route_graph.graph['routes'] = {self.id: self._surrender_to_graph()}
+        route_graph.graph['services'] = {}
+        route_graph.graph['change_log'] = change_log.ChangeLog()
         return route_graph
 
+    def add_additional_attributes(self, attribs: dict):
```
KasiaKoz, line 388: necessary as users can now add any attributes they like to routes, services and stops

genet/schedule_elements.py
```diff
@@ -792,46 +1051,732 @@ def plot(self, show=True, save=False, output_dir=''):
...
+                lambda x: [(k, use_schedule.sanitise_time(time, gtfs_day)) for k, time in x.items()]).values)
+        df = DataFrame({
+            col: np.repeat(df[col].values, df['trips'].str.len())
+            for col in set(df.columns) - {'trips'}}
+        ).assign(trip=trips[:, 0],
+                 trip_dep_time=trips[:, 1]).sort_values(by=['route', 'trip', 'departure_time']).reset_index(drop=True)
+
+        df['departure_time'] = df['trip_dep_time'] + df['departure_time']
+        df['arrival_time'] = df['trip_dep_time'] + df['arrival_time']
+        df = df.drop('trip_dep_time', axis=1)
```
KasiaKoz, line 1094: something that could have been done in a separate PR.... instead of instantiating all of the services and routes within the Schedule to make a composite of their own trips dataframes, this draws on the new method to generate a dataframe with given keys and expands it to give a DataFrame which gives stop to stop rows with time stamps for all trips.

genet/schedule_elements.py
```diff
@@ -792,46 +1051,732 @@ def plot(self, show=True, save=False, output_dir=''):
...
-        return len([r for id, r in self.routes()])
+        return len(self._graph.graph['routes'])
+
+    def has_stop(self, stop_id):
+        """
+        Returns True if a stop with ID `stop_id` exists in the Schedule, False otherwise
+        """
+        return self._graph.has_node(stop_id)
+
+    def service_attribute_summary(self, data=False):
```
KasiaKoz, line 1152: loads of new methods for Schedule from this point on, notice that they exist on Schedule level only. The only change methods that exist for sub elements (services, routes) are reindex methods. I wonder if it's worth adding those methods to those classes too. I'm tempted to leave it like this and see if the users get tripped up by it, and add them if there is a need later on..

genet/utils/graph_operations.py
```diff
@@ -282,6 +236,76 @@ def get_the_data(attributes, key):
     return data
 
 
+def build_attribute_dataframe(iterator, keys: Union[list, str], index_name: str = None):
```
KasiaKoz, line 239: mostly copy paste of what used to be in a Network method, now reused in a bunch of places

genet/exceptions.py
```diff
@@ -0,0 +1,62 @@
+class ScheduleElementGraphSchemaError(Exception):
```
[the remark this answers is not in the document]
mfitz: `brownie_points += 100`

mfitz (approved):
I don't feel like I can fully comprehend what is happening in this change, so I have only a couple of comments and questions. The changes feel pretty big - do you have any idea of the effect on performance for "typical" (if there is such a thing) use cases?

The test suite is very slow these days (when I ran it just now it took 8 minutes) - do you think we should look into that? Is it indicative of any performance degradation in the code?

From the conversation we had about the notebooks in this repo, it sounds like you will have run each of them manually - can you describe what you did, and what the results were? I think you're right - now is the time to look at creating some kind of automated test for those, and I'm happy to help with that.

genet/core.py
```diff
@@ -188,35 +188,17 @@ def node_attribute_data_under_key(self, key):
...
-            col_series.name = name
-
-            if df is not None:
-                df = df.merge(pd.DataFrame(col_series), left_index=True, right_index=True, how='outer')
-            else:
-                df = pd.DataFrame(col_series)
-        if index_name:
-            df.index = df.index.set_names([index_name])
-        return df
+        return graph_operations.build_attribute_dataframe(self.nodes(), keys=keys, index_name=index_name)
```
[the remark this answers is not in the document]
gac55: Helpful to have the same interface for both

genet/core.py
```diff
@@ -235,31 +217,197 @@ def link_attribute_data_under_key(self, key: Union[str, dict]):
...
+            * any: means at least one condition needs to be met
+
+        :param mixed_dtypes: True by default, used if values under dictionary keys queried are single values or lists of
+        values e.g. as in simplified networks.
+        :return: list of link ids in the network satisfying conditions
+        """
+        return graph_operations.extract_on_attributes(
+            self.links(), conditions=conditions, how=how, mixed_dtypes=mixed_dtypes)
+
+    def links_on_modal_condition(self, modes: Union[str, list]):
```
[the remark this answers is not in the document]
gac55: Laying the foundations for better transparency, great

genet/core.py
```diff
@@ -1253,7 +1392,7 @@ def write_to_matsim(self, output_dir):
         matsim_xml_writer.write_matsim_network(output_dir, self)
         if self.schedule:
             self.schedule.write_to_matsim(output_dir)
-        self.change_log.export(os.path.join(output_dir, 'change_log.csv'))
+        self.change_log.export(os.path.join(output_dir, 'network_change_log.csv'))
```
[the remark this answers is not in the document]
gac55: I like that. They're two separate things and it makes sense to have them distinct

genet/schedule_elements.py
gac55: Excellent, some human readable attributes for post processing will be very handy

genet/schedule_elements.py
gac55: Sounds reasonable. I can't quite go through the process in my head, seems reasonable to wait and have some user feedback

gac55 (approved):

genet/modify/change_log.py
```diff
@@ -4,7 +4,7 @@
 from typing import Union, List
 
 
-class ChangeLog:
+class ChangeLog(pd.DataFrame):
```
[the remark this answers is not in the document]
KasiaKoz: so when I started genet I knew I wanted to log changes but I wasn't sure how to do it. I settled on a dataframe because it's simple and most people know how to use pandas. At this point I don't think the changelog will ever not be a dataframe, it works so there is no point changing it. The only thing that might change is a couple of columns might get deleted because of redundancy one day. The `ChangeLog` had literally just one attribute `log` and that was to access the dataframe. I think it makes sense for it to be a dataframe, but I do get your point that it makes it difficult to change it. Users don't really interact with this object tbh, they can view it but it's mostly in the background for insurance

KasiaKoz:
> do you have any idea of the effect on performance for "typical" (if there is such a thing) use cases?

not really, I think typical use cases could be based on the pre-baked scripts and a performance benchmark worked out. I think right now the matsim schedule and gtfs read methods are comparatively quite slow, but these changes make it closer to speeding them up.

> The test suite is very slow these days (when I ran it just now it took 8 minutes) - do you think we should look into that? Is it indicative of any performance degradation in the code?

I thinking looking into that would be awesome. It's been slow for a while though and I keep adding more tests so it's hard to say if it's because of degradation or I just wrote the tests stupidly. I think the tests could be cleaned up in general to make them easier to follow (I grab fixtures from different modules or use the same names and it's not cool...)

>From the conversation we had about the notebooks in this repo, it sounds like you will have run each of them manually - can you describe what you did, and what the results were? I think you're right - now is the time to look at creating some kind of automated test for those, and I'm happy to help with that.

I only ran the two I was changing 🙈  I think I will re-run the whole suite once I'm done with working on the schedule-related tasks, because it's a lot of faff and jupyter changes also mean wiki changes... I base the wiki pages largely on the notebooks 🙈  So, what I think would be useful:
1. automated run of jupyter notbeooks to check code changes don't break anything
2. automated run of pre-baked scripts, in a similar vibe to notebooks---not to check outputs of the scripts, just that they work and *maybe additionally* create outputs as expected, so e.g. I expect a `network.xml` and `network_chenge_log.csv` files, but I don't care about what's inside those files, unit tests check those things. Checking those scripts is probably more important, we use them with trash pandas to simplify networks, reproject and so on. (but maybe matesto already covers this?)
3. automated jupyter notebooks -> markdown and auto update of the related wiki pages upon merge to master - the dream 😂

PR #62, opened 2021-03-01
Lab 1014 NZ overflow

KasiaKoz:
Some of the S2 indices in NZ are too large to be saved as integers. This PR adds a step to convert to strings if that's the case.

During this work weird pyproj behaviour was fixed too. Transformers were missing the `always_xy=True` which resulted in NZ lat lon values being switched. Now all projections to `epsg:4326` return `lon, lat` (`= x, y`) regardless of where they are in the world. 

Back to the bug:
Error stack trace:
```
2021-02-15 21:18:22,199 - Generating geojson outputs for the entire network in /mnt/efs/network/20210212_puma_simplify_updated/simplify_output/standard_outputs
/home/arup/genet-venv/lib/python3.8/site-packages/pyproj/crs/crs.py:53: FutureWarning: '+init=<authority>:<code>' syntax is deprecated. '<authority>:<code>' is the preferred initialization method. When making the change, be mindful of axis order changes: https://pyproj4.github.io/pyproj/stable/gotchas.html#axis-order-changes-in-proj-6
  return _prepare_from_string(" ".join(pjargs))
2021-02-15 21:19:25,783 - Saving network graph nodes and links geojsons to /mnt/efs/network/20210212_puma_simplify_updated/simplify_output/standard_outputs
Traceback (most recent call last):
  File "scripts/simplify_network.py", line 78, in <module>
    n.generate_standard_outputs(os.path.join(output_dir, 'standard_outputs'))
  File "/home/arup/genet/genet/core.py", line 1183, in generate_standard_outputs
    geojson.generate_standard_outputs(self, output_dir, gtfs_day)
  File "/home/arup/genet/genet/outputs_handler/geojson.py", line 145, in generate_standard_outputs
    save_network_to_geojson(n, output_dir)
  File "/home/arup/genet/genet/outputs_handler/geojson.py", line 53, in save_network_to_geojson
    save_geodataframe(graph_nodes, 'network_nodes', output_dir)
  File "/home/arup/genet/genet/outputs_handler/geojson.py", line 41, in save_geodataframe
    gdf.to_file(os.path.join(output_dir, f'{filename}.geojson'), driver='GeoJSON')
  File "/home/arup/genet-venv/lib/python3.8/site-packages/geopandas/geodataframe.py", line 533, in to_file
    to_file(self, filename, driver, schema, index, **kwargs)
  File "/home/arup/genet-venv/lib/python3.8/site-packages/geopandas/io/file.py", line 193, in to_file
    colxn.writerecords(df.iterfeatures())
  File "/home/arup/genet-venv/lib/python3.8/site-packages/fiona/collection.py", line 342, in writerecords
    self.session.writerecs(records, self)
  File "fiona/ogrext.pyx", line 1195, in fiona.ogrext.WritingSession.writerecs
  File "fiona/ogrext.pyx", line 366, in fiona.ogrext.OGRFeatureBuilder.build
OverflowError: Python int too large to convert to C long
```
Relevant test before and after:
Screenshot 2021-03-01 at 11 10 59
Screenshot 2021-03-01 at 11 20 50


genet/inputs_handler/osm_reader.py
```diff
@@ -116,16 +116,16 @@ def create_s2_indexed_osm_graph(response_jsons, config, num_processes, bidirecti
 
 
 def generate_graph_nodes(nodes, epsg):
-    input_to_output_transformer = Transformer.from_crs('epsg:4326', epsg)
+    input_to_output_transformer = Transformer.from_crs('epsg:4326', epsg, always_xy=True)
```
mfitz, line 119: Would we ever want `always_xy=False` (which is the default, I guess)? Wondering if we should create our own wrapper around this in the form of a factory function or suchlike.

genet/outputs_handler/sanitiser.py
```diff
@@ -18,6 +20,9 @@ def sanitise_geodataframe(gdf):
             gdf[col] = gdf[col].apply(lambda x: ','.join(x))
         elif gdf[col].apply(lambda x: isinstance(x, dict)).any():
             gdf[col] = gdf[col].apply(lambda x: str(x))
+    for col in gdf.select_dtypes(include=number).columns.tolist():
```
mfitz, line 23: Does changing the type to a string have any impact on the data in its serialised form? I had a look at generated GeoJSON in Matesto for test town, and I see:

```json
    {
      "type": "Feature",
      "properties": {
        "index": "7",
        "x": 500,
        "y": 3500,
        "id": "7",
        "lon": -7.553841327916719,
        "lat": 49.79847109699483,
        "s2_id": 5206348332324383000
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          -7.553841327916719,
          49.79847109699483
        ]
      }
    }
```

Will this mean `"s2_id": 5206348332324383000` becomes `"s2_id": "5206348332324383000"`? If so, are there any implications for processes that read these GeoJSON files?

mfitz (approved):

genet/inputs_handler/osm_reader.py
KasiaKoz: No idea. I tried to see why you would ever set this as False or feel the need to swap the order of points and I just cannot understand it.

genet/outputs_handler/sanitiser.py
KasiaKoz: They only become strings if they're too large, so the test town indices should remain as they are. Nothing apart from kepler or GIS software would be reading these geojson files, I haven't tested it but I think kepler will be ok with those strings.

PR #70, opened 2021-03-24
use sets in route and service indexing on schedule graph

KasiaKoz:
Fixes a brain fart that had gone too far.
Nodes and edges of a schedule object have `'routes'` and `'services'` data saved on them to be able to extract subgraphs given a route or a service. There is no need for this to be a `list` of service or route IDs, and because lists are ordered, it makes it annoying to test when changes are made that would disrupt the order. We don't care about the order, and those IDs should only feature once per node or edge, so it makes more sense to use sets. Especially as there are instances where we convert those lists to sets to perform an operation and make it a list again.... >_>


mfitz (approved):
Looks good. I do enjoy a good set. Just a lovely, lovely data structure :-)

PR #80, opened 2021-04-29
Fix scripts

KasiaKoz:
Fixes broken scripts post https://github.com/arup-group/genet/pull/73 as was done for the simplification script here: https://github.com/arup-group/genet/pull/79

It would be awesome to have automated tests for these scripts: [LAB-1185](https://arupdigital.atlassian.net/browse/LAB-1185) records that. I included example invocation for the scripts in that card (because I had to run them post fixing to check they work :D)

There is only one issue with the `send_google_directions_requests_for_network.py` which will not work properly unless we give it an api key. And I'm not sure how to fake it :D

### DIFF
--- a/scripts/generate_standard_outputs.py
+++ b/scripts/generate_standard_outputs.py
@@ -1,14 +1,12 @@
 import argparse
-import genet as gn
 import logging
-import time
 import os
-import json
+
+from genet import read_matsim
 from genet.utils.persistence import ensure_dir
 
-
 if __name__ == '__main__':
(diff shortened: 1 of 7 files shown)


mfitz (approved):
Looks good. I will pick up https://arupdigital.atlassian.net/browse/LAB-1185 at some point soon, unless you want to do that yourself. I'll have a think about what to do about the Google Maps API key - maybe we can monkey patch or mock that in some way in a smoke test.

PR #84, opened 2021-07-30
Lab 1226 Refine Road Pricing

KasiaKoz:
Adds a `Cordon` class to hold information about tolling on network links.
All previous road pricing methods are retained and wraps added to have them work with this new class.
It's all pretty basic but it introduces a way to represent cordon tolls in memory that isn't dependent on OSM ids and a mapping between them and a network object.


genet/use/road_pricing.py
```diff
@@ -1,27 +1,99 @@
...
 from lxml import etree as et
 from lxml.etree import Element, SubElement, Comment
 from tqdm import tqdm
 
 
-def extract_network_id_from_osm_csv(network, attribute_name, osm_csv_path, outpath):
+class Cordon:
+    def __init__(self, df_tolls: pd.DataFrame = None):
+        if df_tolls is None:
+            self.tolls_df = pd.DataFrame(
```
mfitz, line 15: Are we deliberately varying the member name here (`df_tolls` v `tolls_df` on the other side of the `if` clause at line 28)? I can see from the coverage report that line 15 is not tested, so it makes me think maybe this is a typo?

mfitz (approved):
Looks good.

I ran the coverage report and got:

```
675 passed, 1 xfailed, 442 warnings in 570.16s (0:09:30)
```

In separate branches, we should:
- clean up all of those warnings
- see if we can speed those tests up

genet/use/road_pricing.py
KasiaKoz: oops, typo!

PR #85, opened 2021-08-06
Add manifest

mfitz:
Adds a [manifest](https://packaging.python.org/guides/using-manifest-in/) so that genet's source distribution, as pulled in via scop's [requirements.txt](https://github.com/arup-group/scop/blob/finish-csv-read-write/requirements.txt), for example, includes the necessary genet config files from `genet/configs`.

In `scop`, after `pip install -r requirements.txt` but before adding the genet manifest:

Screen Shot 2021-08-06 at 16 36 14

After adding the manifest, the missing genet config files are now available after `pip install -r requirements.txt`:

Screen Shot 2021-08-06 at 16 35 20
Screen Shot 2021-08-06 at 16 35 34

You can now see the genet configs directory inside my scop venv:

```bash
$ tree /Users/mickyfitz/.pyenv/versions/scop-3.7.6/lib/python3.7/site-packages/genet/configs/
/Users/mickyfitz/.pyenv/versions/scop-3.7.6/lib/python3.7/site-packages/genet/configs/
├── OSM
│   ├── default_config.yml
│   └── slim_config.yml
└── vehicles
    └── vehicle_definitions.yml

2 directories, 3 files
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include genet/configs *


KasiaKoz (approved):
yassss

PR #90, opened 2021-08-27
PT network route GeoDataFrame and geojson output

KasiaKoz:
Adds method to generate geodataframes of PT routes which take geometries from network links referenced in their routes.
So we can generate pretty pictures like the one below. Also added to standard outputs, useful for validation. The geodataframe has mode, service and route id columns for easy subsetting, as well as route names for human ease.
market_town_bus_routes_global_routes_backgroud


gac55 (approved):
🚀  very cool

PR #91, opened 2021-08-27
Connect disconnected subgraphs

KasiaKoz:
Adds methods to find and add `2*(n-1)` links for `n` disconnected subgraphs of mode `modes`. 
The added links inherit an average of `'freespeed', 'capacity'` attributes from neighbouring links and the specified mode(s) or a union of modes form neighbouring links. 
`'freespeed', 'capacity'` can be scaled up or down using the `weight` variable.


gac55 (approved):
Am I right in saying this is what was developed as part of the NZ "island" problem? It's good that we had some modelling tests that helped shape these functionality. 

Tests look good too 🚀

PR #94, opened 2021-09-14
Fix runtime warnings

### LINKED ISSUE #93: GeNet produces many warnings
Running the unit test suite produces 453 warnings. Some of these warnings are caused by the same `pyproj` syntax deprecation described by [this recent Puma PR](https://github.com/arup-group/puma/pull/63), but there are also others with different origins. 

Screen Shot 2021-09-09 at 13 45 31

When I run GeNet in Matesto to simplify a test town network, I see six of these pyproj deprecation warnings:

```bash
FutureWarning: '+init=<authority>:<code>' syntax is deprecated. '<authority>:<code>' is the preferred initialization method. When making the change, be mindful of axis order changes: https://pyproj4.github.io/pyproj/stable/gotchas.html#axis-order-changes-in-proj-6
  return _prepare_from_string(" ".join(pjargs))
```

It would be good to fix some of these - at the very least, we should sort out the `pyproj` warnings.

mfitz:
Closes #93 

Fixes a few hundred warnings - still a few hanging around.

I also had to bump pytest-related versions to make the parallel test stuff work on my local machine, and modified the number of processes so that it is automatically matched to the number of available processors reported by the machine.

I had a brief look at the `xfail` test, but couldn't see a quick fix for it.

#### Before
Screen Shot 2021-09-14 at 23 59 43

#### After
Screen Shot 2021-09-15 at 00 00 15

### DIFF
--- a/genet/core.py
+++ b/genet/core.py
@@ -188,7 +188,8 @@ class Network:
             e.g. {'attributes': {'osm:way:name': 'text'}}
         :return: pandas.Series
         """
-        return pd.Series(graph_operations.get_attribute_data_under_key(self.nodes(), key))
+        data = graph_operations.get_attribute_data_under_key(self.nodes(), key)
+        return pd.Series(data, dtype=graph_operations.get_pandas_dtype(data))
 
     def node_attribute_data_under_keys(self, keys: Union[list, set], index_name=None):
         """
--- a/genet/core.py
+++ b/genet/core.py
@@ -20,6 +20,7 @@ import genet.outputs_handler.matsim_xml_writer as matsim_xml_writer
 import genet.outputs_handler.sanitiser as sanitiser
 import genet.schedule_elements as schedule_elements
 import genet.utils.dict_support as dict_support
+import genet.utils.pandas_helpers as pd_helpers
 import genet.utils.graph_operations as graph_operations
 import genet.utils.parallel as parallel
 import genet.utils.persistence as persistence
@@ -189,7 +190,7 @@ class Network:
         :return: pandas.Series
         """
         data = graph_operations.get_attribute_data_under_key(self.nodes(), key)
-        return pd.Series(data, dtype=graph_operations.get_pandas_dtype(data))
+        return pd.Series(data, dtype=pd_helpers.get_pandas_dtype(data))
 
     def node_attribute_data_under_keys(self, keys: Union[list, set], index_name=None):
         """
@@ -622,7 +623,7 @@ class Network:
         # end with updated links_and_attributes dict
         add_to_link_id_mapping = df_links[['from', 'to', 'multi_edge_idx']].T.to_dict()
         df_links = df_links.drop('multi_edge_idx', axis=1)
-        links_and_attributes = {_id: {k: v for k, v in m.items() if dict_support.notna(v)} for _id, m in
+        links_and_attributes = {_id: {k: v for k, v in m.items() if pd_helpers.notna(v)} for _id, m in
                                 df_links.T.to_dict().items()}
 
         # update link_id_mapping
--- a/genet/core.py
+++ b/genet/core.py
@@ -144,8 +144,7 @@ class Network:
         self.apply_attributes_to_nodes(new_nodes_attribs)
 
         # reproject geometries
-        gdf_geometries = gpd.GeoDataFrame(self.link_attribute_data_under_keys(['geometry']))
-        gdf_geometries.crs = self.epsg
+        gdf_geometries = gpd.GeoDataFrame(self.link_attribute_data_under_keys(['geometry']), crs=self.epsg)
         gdf_geometries = gdf_geometries.to_crs(new_epsg)
         new_link_attribs = gdf_geometries.T.to_dict()
         self.apply_attributes_to_links(new_link_attribs)
--- a/genet/core.py
+++ b/genet/core.py
@@ -35,7 +35,7 @@ class Network:
     def __init__(self, epsg):
         self.epsg = epsg
         self.transformer = Transformer.from_crs(epsg, 'epsg:4326', always_xy=True)
-        self.graph = nx.MultiDiGraph(name='Network graph', crs={'init': self.epsg}, simplified=False)
+        self.graph = nx.MultiDiGraph(name='Network graph', crs=self.epsg, simplified=False)
         self.schedule = schedule_elements.Schedule(epsg)
         self.change_log = change_log.ChangeLog()
         self.spatial_tree = spatial.SpatialTree()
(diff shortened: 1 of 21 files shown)


genet/utils/graph_operations.py
```diff
@@ -270,6 +272,18 @@ def build_attribute_dataframe(iterator, keys: Union[list, str], index_name: str
     return df
 
 
+def get_pandas_dtype(dict):
```
mfitz, line 275: As discussed - this feels pretty hacky, but it gets the job done and only broke a single test in the end. I'm still hoping for a more elegant solution, but at least we know this does work.

requirements.txt
```diff
@@ -40,7 +40,7 @@ numpy>=1.18.4
 osmnx==0.15.0
 osmread @ git+https://github.com/dezhin/osmread.git@d8d3fe5edd15fdab9526ea7a100ee6c796315663#egg=osmread-0.2.dev0
 packaging==20.4
-pandas>=1.0.3
+pandas==1.3.3
```
mfitz, line 43: Upgrading pandas scared me slightly, but it didn't break any unit tests or notebooks.

requirements.txt
```diff
@@ -54,8 +54,8 @@ py>=1.8.1
 Pygments==2.7.4
 pyparsing==2.4.7
 pyproj>=3.1.0
-pytest>=5.4.2
-pytest-cov>=2.8.1
+pytest==6.2.5
```
mfitz, line 57: Had to upgrade these to make the parallel test stuff work on my machine.

tests/test_core_schedule.py
```diff
@@ -1511,15 +1508,16 @@ def test_transforming_schedule_to_gtfs(schedule):
...
          'platform_code': {'5': float('nan'), '6': float('nan'), '7': float('nan'), '8': float('nan'),
                            '3': float('nan'), '2': float('nan'), '4': float('nan'), '1': float('nan')}}
-    )
-    assert_semantically_equal(
-        gtfs['routes'].to_dict(),
-        {'route_id': {0: 'service'}, 'route_short_name': {0: 'name_2'}, 'route_long_name': {0: ''},
-         'agency_id': {0: float('nan')}, 'route_desc': {0: float('nan')}, 'route_url': {0: float('nan')},
+    actual_stops = gtfs['stops'].to_dict()
+    assert_semantically_equal(expected_stops, actual_stops)
+    expected_routes = {'route_id': {0: 'service'}, 'route_short_name': {0: 'name_2'}, 'route_long_name': {0: ''},
```
mfitz, line 1513: Strangely, I only had to replace NaN with None in expected routes, and not the other GTFS-related dictionaries. Not sure why this is the case, and did not investigate. Can you think of a reason, and does this seem like a problem?

tests/test_data/matsim/network.xml
```diff
@@ -2,7 +2,7 @@
 <!DOCTYPE network SYSTEM "http://www.matsim.org/files/dtd/network_v2.dtd">
 <network>
 	<attributes>
-		<attribute name="crs" class="java.lang.String">{'init': 'epsg:27700'}</attribute>
+		<attribute name="crs" class="java.lang.String">epsg:27700</attribute>
```
mfitz, line 5: Wasn't sure of the implications of having to make change to a test network - could this mean existing networks will be broken now? I used the branch of genet for network simplification in a Matesto smoketest pipeline, and nothing fell over, but that doesn't feel like comprehensive reassurance. Presumably any network written by an older version of GeNet (or Puma?) will include the CRS attribute in the deprecated style?

genet/core.py
```diff
@@ -144,16 +144,15 @@ def reproject(self, new_epsg, processes=1):
         self.apply_attributes_to_nodes(new_nodes_attribs)
 
         # reproject geometries
-        gdf_geometries = gpd.GeoDataFrame(self.link_attribute_data_under_keys(['geometry']))
-        gdf_geometries.crs = self.epsg
+        gdf_geometries = gpd.GeoDataFrame(self.link_attribute_data_under_keys(['geometry']), crs=self.epsg)
```
KasiaKoz, line 147: nice

genet/utils/graph_operations.py
```diff
@@ -213,7 +214,7 @@ def get_attribute_data_under_key(iterator: Iterable, key: Union[str, dict]):
     :param iterator: list or iterator yielding (index, attribute_dictionary)
     :param key: either a string e.g. 'modes', or if accessing nested information, a dictionary
         e.g. {'attributes': {'osm:way:name': 'text'}}
-    :return: dictionary where keys are indicies and values are data stored under the key
+    :return: dictionary where keys are indices and values are data stored under the key
```
KasiaKoz, line 217: I always do that typo, always

genet/utils/graph_operations.py
KasiaKoz: I think it's fine. I wonder where we should put this method, there is another one here: https://github.com/arup-group/genet/blob/master/genet/utils/dict_support.py#L115 which is also in theme of working with pandas data, and `dict_support` is not really the best place for it... should we maybe start a new module in utils.. ?

requirements.txt
KasiaKoz: I changed it to >= quite recently, I wanted it to play nicely with MC (but don't need to in the end because SCOP went a different direction). I prefer the fixed versions I think

tests/test_core_schedule.py
KasiaKoz: huh, weird! I would have thought that None would happen when you have a mixture of empty and non empty values in a column, but that's not the case here. I don't think it should make a difference though, pandas considers None and NaN empty, so if you do `fillna` both None and NaN values will be considered

tests/test_data/matsim/network.xml
KasiaKoz: I don't think this will break anything actually, lol
GeNet just writes the projection to xml, doesnt use it (user has to specify the projection for the network always), it does use simplified tag though https://github.com/arup-group/genet/blob/909e03422c673db83e987cf7fcaeb0632c435078/genet/inputs_handler/read.py#L47  which gets read from xml here: https://github.com/arup-group/genet/blob/909e03422c673db83e987cf7fcaeb0632c435078/genet/inputs_handler/matsim_reader.py#L161 

I think it would be cool to try read the projection from xml, but we've been doing fine with people specifying each time so I don't feel a strong pull towards it

KasiaKoz (approved):
❤️ the new warning numbers
And loving that we no longer have this dictionary 'init' BS all around, thanks @mfitz !

genet/utils/graph_operations.py
mfitz: Yes, I also wondered if this was the best place for it. Perhaps a new `pandas-utils` module in utils is the thing?

tests/test_data/matsim/network.xml
mfitz: Ah, okay. Yes, reading the projection from XML seems like a nice-to-have at best.

PR #97, opened 2021-10-21
Lab 1315 kepler viz

KasiaKoz:
[LAB-1315](https://arupdigital.atlassian.net/browse/LAB-1315)

Replaces old matplotlib plots generated using `osmnx` with dynamic kepler plots. These can be visualised in a jupyter notebook and/or saved to html.

The kepler maps use custom configs so that:
- colour schemes are consistent
- labels of cities are not on the map
- Schedule plots have stop names plastered on the map

But the main kepler plot method can be used with any geodataframe without a config, which will mean a standards kepler experience, and can be saved to html at current state (so after you set up what you want to visualise and how).

### DIFF
--- a/genet/schedule_elements.py
+++ b/genet/schedule_elements.py
@@ -197,7 +197,7 @@ class ScheduleElement:
 
     def kepler_map(self, output_dir='', file_name='kepler_map', data=False):
         gdf = self.to_geodataframe()
-        if data != True:
+        if data is not True:
             gdf['links'] = sanitiser._subset_plot_gdf(data, gdf['links'], base_keys={'u', 'v', 'geometry'})
             gdf['nodes'] = sanitiser._subset_plot_gdf(data, gdf['nodes'], base_keys={'id', 'lat', 'lon'})
 
--- a/genet/core.py
+++ b/genet/core.py
@@ -108,7 +108,7 @@ class Network:
         network_links = self.to_geodataframe()['links']
         schedule_routes = self.schedule_network_routes_geodataframe()
 
-        if data != True:
+        if data is not True:
             network_links = sanitiser._subset_plot_gdf(data, network_links, base_keys={'id', 'geometry'})
             schedule_routes = sanitiser._subset_plot_gdf(data, schedule_routes, base_keys={'route_id', 'geometry'})
 
@@ -134,7 +134,7 @@ class Network:
         """
         network_links = self.to_geodataframe()['links']
 
-        if data != True:
+        if data is not True:
             network_links = sanitiser._subset_plot_gdf(data, network_links, base_keys={'id', 'geometry'})
 
         m = plot.plot_geodataframes_on_kepler_map(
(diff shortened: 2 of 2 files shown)


genet/schedule_elements.py
```diff
@@ -494,16 +511,18 @@ def info(self):
...
-            )
+    def plot(self, output_dir=''):
+        """
+        Plots the route on kepler map.
+        Ensure all prerequisites are installed https://docs.kepler.gl/docs/keplergl-jupyter#install
+        :param output_dir: output directory for the image, if passed, will save plot to html
+        :return:
+        """
+        m = self.kepler_map()
+        if output_dir:
```
mfitz, line 522: This same block of code appears in several places, where the only difference seems to be the file name. Maybe we can refactor this into a little function?

genet/utils/plot.py
```diff
@@ -1,45 +1,293 @@
...
+    if kepler_config in KEPLER_CONFIGS:
+        kepler_config = KEPLER_CONFIGS[kepler_config]
+    elif isinstance(kepler_config, dict):
+        kepler_config = kepler_config
     else:
-        filepath = None
-    return ox.plot_graph(
-        g, filepath=filepath, node_size=2, edge_color=e_c, save=save, show=show)
+        kepler_config = {}
+    m = keplergl.KeplerGl(data=gdfs, height=height, config=kepler_config)
```
mfitz, line 22: The `m` variable isn't used - an opportunity to delete a line :-P

tests/test_core_components_route.py
```diff
@@ -107,9 +108,19 @@ def test_info_shows_id_name_and_len_of_stops_and_trips(route):
...
+    mocker.patch.object(plot, 'plot_geodataframes_on_kepler_map')
     route.plot()
-    plot.plot_graph.assert_called_once()
+    plot.plot_geodataframes_on_kepler_map.assert_called_once()
+
+
+def test_plot_saves_to_the_specified_directory(tmpdir, route):
+    filename = f'route_{route.id}_map'
+    expected_plot_path = os.path.join(tmpdir, filename+'.html')
+    assert not os.path.exists(expected_plot_path)
```
mfitz, line 119: I like the negative assertions before calling the code being tested 👍

mfitz (approved):
Looks good. 

I'm wondering about the prerequisites for the Kepler module - should they be mentioned in the README or somewhere on the wiki?

Not related to this PR, but while browsing the wiki I noticed a couple of broken image links, for example [in this section](https://github.com/arup-group/genet/wiki/Usage:-1.-Intro#network-data-structure), which points at [this URL](https://github.com/arup-group/genet/wiki/images/genet_network_graph.png), which I guess should be [this](https://github.com/arup-group/genet/blob/master/images/genet_network_graph.png). The relative path needs fixing up. Not for this PR, but I mention it because I'll likely forget it otherwise :-)

PR #99, opened 2021-10-28
Osm read fix

KasiaKoz:
Noticed that the osm read method was not making use of multiprocessing 🤦‍♀️ 
Also added a flag to ignore adding these changes to `changelog`, which is useful for keeping a log of differences between networks but bloated and not very useful when creating a network from OSM.


ana-kop (approved):
Looks great to me! 👍

PR #102, opened 2021-11-30
PT headways

KasiaKoz:
MATSim requires explicit trips with vehicles  for all PT Services​. Sometimes we only know the headway in a time bracket for a service, this PR adds a few new convenience methods to:

- Update existing trips using headway information [[ref for a Route object](https://github.com/arup-group/genet/compare/pt-headways?expand=1#diff-143a6df195b3c819f70dcf252a26f450e5010756d4e6df55bc79c092e5e08654R673)] [[ref for Route called within a Schedule object](https://github.com/arup-group/genet/compare/pt-headways?expand=1#diff-143a6df195b3c819f70dcf252a26f450e5010756d4e6df55bc79c092e5e08654R1454)]
- Create new Services using headway information [[ref](https://github.com/arup-group/genet/compare/pt-headways?expand=1#diff-143a6df195b3c819f70dcf252a26f450e5010756d4e6df55bc79c092e5e08654R406)]
- Compute headway information between each trip for each route and present it as a `pandas.DataFrame` [[ref](https://github.com/arup-group/genet/compare/pt-headways?expand=1#diff-143a6df195b3c819f70dcf252a26f450e5010756d4e6df55bc79c092e5e08654R1398)]
- Compute headway stats summary (mean, min, max, etc) for each route and present it as a `pandas.DataFrame` [[ref](https://github.com/arup-group/genet/compare/pt-headways?expand=1#diff-143a6df195b3c819f70dcf252a26f450e5010756d4e6df55bc79c092e5e08654R1472)]

This will help with generating PT scenarios.

The headway specification is a dictionary in this format: `(time_from, time_to): headway in minutes`, e.g.​
```python
{('01:00:00', '02:00:00'): 20, ​
 ('02:00:00', '03:00:00'): 30}​
```

Will come out with these trips:​
```python
['route_ID_01:00:00', 
 'route_ID_01:20:00', 
 'route_ID_01:40:00', 
 'route_ID_02:00:00', 
 'route_ID_02:30:00', 
 'route_ID_03:00:00']
```

This can be more complicated with only some trips being injected among an existing set of trips to try and match a headway, but we propose this simple addition to current functionality for the time being.


gac55 (commented):
Clarification question.

Does this implement a given headway across the entire day uniformly? I can imagine this having unintended consequences, for example, creating services early in the morning or late at night. Does this make that configurable?

gac55 (approved):
Excellent, understood now. 🚀

PR #103, opened 2021-12-02
Sub network/schedule

KasiaKoz:
Methods to return a new `Network` or `Schedule` object subsetted using link IDs and service IDs. 
- Includes a convenience method for spatial subsetting.
- Ensures the network links for services given are retained in the Network and that the resulting graph is strongly connected for the routable agent modes.


genet/core.py
```diff
@@ -420,6 +420,10 @@ def links_on_spatial_condition(self, region_input, how='intersect'):
              the tokens string copied using http://s2.sidewalklabs.com/regioncoverer/
              e.g. '89c25985,89c25987,89c2598c,89c25994,89c25999ffc,89c2599b,89c259ec,89c259f4,89c25a1c,89c25a24'
             - shapely.geometry object, e.g. Polygon or a shapely.geometry.GeometryCollection of such objects
+        :param how:
+            - 'intersect' default, will return IDs of the Services whose at least one Stop intersects the
```
gac55, line 424: Great, I can see use cases for both types selection

genet/core.py
```diff
@@ -432,6 +436,100 @@ def links_on_spatial_condition(self, region_input, how='intersect'):
...
+    def subnetwork(self, links: Union[list, set], services: Union[list, set] = None):
+        """
+        Subset a Network object using a collection of link IDs and (optionally) service IDs
+        :param links: Link IDs to be retained in the new Network
+        :param services: optional, collection of service IDs in the Schedule for subsetting.
+        :return: A new Network object that is a subset of the original
+        """
+        logging.info('Subsetting a Network will likely result in a disconnected network graph. A cleaner will be ran '
+                     'that will remove links to make the resulting Network strongly connected for modes: '
+                     'car, walk, bike.')
```
gac55, line 448: Should this be parametrised for future compatibility? Perhaps not, as car, walk, bike is fairly standard and I can't think of any non PT modes we may introduce in the future. I suppose networks for scooters etc is technically possible.

genet/core.py
```diff
@@ -432,6 +436,100 @@ def links_on_spatial_condition(self, region_input, how='intersect'):
...
+
+        logging.info('Subsetted Network is ready - do not forget to validate and visualise your subset!')
+        return subnetwork
+
+    def subnetwork_on_spatial_condition(self, region_input, how='intersect'):
+        """
+        Subset a Network object using a spatial bound
+        :param region_input:
+            - path to a geojson file, can have multiple features
+            - string with comma separated hex tokens of Google's S2 geometry, a region can be covered with cells and
```
gac55, line 475: Lovely to see so many options. I imagine geojson will be v common, but S2 support is 🆒

gac55 (approved):
🚀

PR #107, opened 2022-01-31
Fix loopy simplification

KasiaKoz:
## Problem

During the simplification process, some links get simplified to loops. In cases of teleported PT, this results in undesirable routes.

For example, given a rail stop sequence: 

```
[A, B, C, D]
```

And the expected, teleported, route is: 

```
[A-A, A-B, B-B, B-C, C-C, C-D, D-D]
```

Right now GeNet oversimplifies the end points resulting in long loops, the route is:

```
[S1, B-B, B-C, C-C, S2]
```

the geometry is retained so they look like lines, but are actually very big loop links, i.e.

```
S1: 
   from node: B 
   to node: B 
```

Not tested, but the likely MATSim consequences are agents not being able to access the service at stops A and D, as MATSim does not use our complex geometry (it's just an additional attribute)

## Fix

Change to defining 'end-points' during simplification - those are the nodes retained in the network during the process. We retain nodes that are self-loops to begin with, thus protecting our PT stops from simplifying. This results in a small increase in retained nodes and fewer links being simplified. 

Difference in the number of links being simplified in the test network:
Before: `6838`
After: `6829`

## Bonus

Now, as default, all links that were simplified to self-loops, will be removed (as they are useless) unless used by PT (we use small self loops as PT stops) or an additional param is given to retain them.

This does not affect the connectivity of the graph.

## Matesto pass

Screenshot 2022-02-07 at 18 09 22
Screenshot 2022-02-07 at 18 09 29
Screenshot 2022-02-07 at 18 09 35

### DIFF
--- a/.github/workflows/build_pipeline.yml
+++ b/.github/workflows/build_pipeline.yml
@@ -32,7 +32,7 @@ jobs:
         pip install -e .
     - name: Lint with flake8
       run: |
-        flake8 . --max-line-length 120 --count  --show-source --statistics --exclude=scripts,tests,notebooks
+        flake8 . --max-line-length 120 --count  --show-source --statistics --exclude=scripts,tests,notebooks,venv
     - name: Unit tests
       run: |
         export CURL_CA_BUNDLE=/etc/ssl/certs/ca-certificates.crt
--- a/genet/core.py
+++ b/genet/core.py
@@ -229,11 +229,12 @@ class Network:
         pt_stop_loops = set(self.schedule.stop_attribute_data(keys=['linkRefId'])['linkRefId'])
         to_remove = loops - pt_stop_loops
         if to_remove:
-            logging.info(f'Simplification led to {len(loops)} in the network. {len(to_remove)} are not connected'
-                         'to PT stops.')
+            logging.info(f'Simplification led to {len(loops)} self-loop links in the network. '
+                         f'{len(to_remove)} are not connected to the PT stops.')
             if not keep_loops:
-                logging.info('These loops will now be removed. To disable this behaviour, use `keep_loops=True`.'
-                             'Investigate the change log for more information about these links')
+                logging.info('The self-loops with no reference to PT stops will now be removed. '
+                             'To disable this behaviour, use `keep_loops=True`. '
+                             'Investigate the change log for more information about these links.')
                 self.remove_links(to_remove)
 
         # mark graph as having been simplified
--- a/genet/core.py
+++ b/genet/core.py
@@ -217,7 +217,7 @@ class Network:
             and 1 process is often preferable --- there is overhead for splitting and joining the data.
         :param keep_loops: bool, defaults to False. Simplification often leads to self-loops, these will be removed
             unless keep_loops=True
-        :return: set of self-loop link IDs
+        :return: None, updates Network object
         """
         if self.is_simplified():
             raise RuntimeError('This network has already been simplified. You cannot simplify the graph twice.')
--- a/genet/core.py
+++ b/genet/core.py
@@ -215,8 +215,9 @@ class Network:
         Simplifies network graph, retaining only nodes that are junctions
         :param no_processes: Number of processes to split some computation across. The method is pretty fast though
             and 1 process is often preferable --- there is overhead for splitting and joining the data.
-        :param keep_loops: bool, simplification often leads to
-        :return:
+        :param keep_loops: bool, defaults to False. Simplification often leads to self-loops, these will be removed
+            unless keep_loops=True
+        :return: set of self-loop link IDs
         """
         if self.is_simplified():
             raise RuntimeError('This network has already been simplified. You cannot simplify the graph twice.')
@@ -227,20 +228,20 @@ class Network:
         loops = set(df.index)
         # pt stops can be loops
         pt_stop_loops = set(self.schedule.stop_attribute_data(keys=['linkRefId'])['linkRefId'])
-        to_remove = loops - pt_stop_loops
-        if to_remove:
-            logging.info(f'Simplification led to {len(loops)} self-loop links in the network. '
-                         f'{len(to_remove)} are not connected to the PT stops.')
+        useless_self_loops = loops - pt_stop_loops
+        if useless_self_loops:
+            logging.warning(f'Simplification led to {len(loops)} self-loop links in the network. '
+                         f'{len(useless_self_loops)} are not connected to the PT stops.')
             if not keep_loops:
                 logging.info('The self-loops with no reference to PT stops will now be removed. '
                              'To disable this behaviour, use `keep_loops=True`. '
                              'Investigate the change log for more information about these links.')
-                self.remove_links(to_remove)
+                self.remove_links(useless_self_loops)
 
         # mark graph as having been simplified
         self.graph.graph["simplified"] = True
 
-        return to_remove
+        return useless_self_loops
 
     def is_simplified(self):
         return self.graph.graph["simplified"]
--- a/genet/core.py
+++ b/genet/core.py
@@ -237,11 +237,12 @@ class Network:
                              'To disable this behaviour, use `keep_loops=True`. '
                              'Investigate the change log for more information about these links.')
                 self.remove_links(useless_self_loops)
+                # delete removed links from the simplification map
+                self.link_simplification_map = {k: v for k, v in self.link_simplification_map.items() if
+                                                v not in useless_self_loops}
 
         # mark graph as having been simplified
         self.graph.graph["simplified"] = True
-
-        return useless_self_loops
 
     def is_simplified(self):
         return self.graph.graph["simplified"]
--- a/genet/core.py
+++ b/genet/core.py
@@ -231,7 +231,7 @@ class Network:
         useless_self_loops = loops - pt_stop_loops
         if useless_self_loops:
             logging.warning(f'Simplification led to {len(loops)} self-loop links in the network. '
-                         f'{len(useless_self_loops)} are not connected to the PT stops.')
+                            f'{len(useless_self_loops)} are not connected to the PT stops.')
             if not keep_loops:
                 logging.info('The self-loops with no reference to PT stops will now be removed. '
                              'To disable this behaviour, use `keep_loops=True`. '
--- a/genet/core.py
+++ b/genet/core.py
@@ -210,12 +210,36 @@ class Network:
         else:
             self.transformer = None
 
-    def simplify(self, no_processes=1):
+    def simplify(self, no_processes=1, keep_loops=False):
+        """
+        Simplifies network graph, retaining only nodes that are junctions
+        :param no_processes: Number of processes to split some computation across. The method is pretty fast though
+            and 1 process is often preferable --- there is overhead for splitting and joining the data.
+        :param keep_loops: bool, simplification often leads to
+        :return:
+        """
         if self.is_simplified():
             raise RuntimeError('This network has already been simplified. You cannot simplify the graph twice.')
         simplification.simplify_graph(self, no_processes)
+
+        df = self.link_attribute_data_under_keys(keys=['from', 'to'])
+        df = df[df['from'] == df['to']]
+        loops = set(df.index)
+        # pt stops can be loops
+        pt_stop_loops = set(self.schedule.stop_attribute_data(keys=['linkRefId'])['linkRefId'])
+        to_remove = loops - pt_stop_loops
+        if to_remove:
+            logging.info(f'Simplification led to {len(loops)} in the network. {len(to_remove)} are not connected'
+                         'to PT stops.')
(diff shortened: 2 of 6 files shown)


tests/test_core_network.py
```diff
@@ -500,12 +516,36 @@ def test_simplifing_puma_network_results_in_correct_record_of_removed_links_and_
...
 
 
-def test_simplified_network_saves_to_correct_dtds(tmpdir, network_dtd, schedule_dtd):
-    n = read.read_matsim(path_to_network=puma_network_test_file, epsg='epsg:27700',
-                         path_to_schedule=puma_schedule_test_file)
+def test_simplify_does_not_oversimplify_PT_endpoints(
+        puma_network_pre_simplify, puma_network_post_simplify):
+    n = puma_network_pre_simplify
+
+    stops_at_risk = ['5221390681543854913', '5221390302070799085', '5221390323679791901']
```
mfitz, line 523: Are these PT stop IDs? Maybe they should belong to the fixture, so you have, say, a `puma_network_with_pt_stops` fixture that is the network and a list of the PT stop IDs that belong to the network. That way the implicit knowledge of the network, i.e. what the IDs of the PT stop nodes are, is contained in just one place - the fixture - rather than leaking out into the test (or more than one test).

So the test would look more like (with a helper method `assert_link_length` - maybe there is a better, higher level version of that, like `assert_is_blah`, that hides the detail of `link length == 1` -  `assert_valid_pt_node`, or whatever):

```python
def test_simplify_does_not_oversimplify_PT_endpoints(puma_network_with_pt_stops):
    network, pt_stop_ids = puma_network_with_pt_stops
    for stop in pt_stop_ids:
        assert_link_length(stop, 1, network)
        
    network.simplify()
    
     for stop in pt_stop_ids:
         assert_link_length(stop, 1, network)
```

That looks more like the usual basic structure of a unit test:

- setup
- call code under test
- validate expectations

Also known as "Arrange, Act, Assert"

Or... maybe there is a way you can query the network and discover the PT stops, inside the test, rather than having a pre-prepared list of the PT stop IDs? So then, with a helper method `get_pt_stops`,  something like:

```python
def test_simplify_does_not_oversimplify_PT_endpoints(puma_network_with_pt_stops):
    for stop_id in get_pt_stops(puma_network_with_pt_stops):
        assert_link_length(stop_id, 1, puma_network_with_pt_stops)
        
    puma_network_with_pt_stops.simplify()
    
    for stop_id in get_pt_stops(puma_network_with_pt_stops):
        assert_link_length(stop_id, 1, puma_network_with_pt_stops)
```

tests/test_core_network.py
```diff
@@ -500,12 +516,36 @@ def test_simplifing_puma_network_results_in_correct_record_of_removed_links_and_
...
+
+    deleted_self_loops, n = puma_network_post_simplify
+
+    for s in stops_at_risk:
+        assert n.link(n.schedule.stop(s).linkRefId)['length'] == 1
+
+
+def test_simplify_removes_loops_by_default_but_keeps_pt_stop_loops(
+        puma_network_pre_simplify, puma_network_post_simplify):
+    n = puma_network_pre_simplify
```
mfitz, line 535: Again, the code under test is kind of hidden away in a fixture. Fixtures should just provide data to use for the test, really, they shouldn't be calling the code under test. They're part of the setup. We've also got some implicit knowledge of the network (particular link IDs) lurking in this test - is there a way to stop that knowledge from being spread around like this?

tests/test_core_network.py
```diff
@@ -500,12 +516,36 @@ def test_simplifing_puma_network_results_in_correct_record_of_removed_links_and_
...
+    assert n.schedule.stop('5221390681543854913').linkRefId == '5221390681543854913_5221390681543854913'
+    assert n.has_link('5221390681543854913_5221390681543854913')
+    assert n.link('5221390681543854913_5221390681543854913')['from'] == '5221390681543854913'
+    assert n.link('5221390681543854913_5221390681543854913')['to'] == '5221390681543854913'
 
-    n.simplify()
 
+def test_simplified_network_saves_to_correct_dtds(tmpdir, network_dtd, schedule_dtd, puma_network_post_simplify):
+    deleted_self_loops, n = puma_network_post_simplify
     n.write_to_matsim(tmpdir)
```
mfitz, line 549: We're kind of testing two things at once here - network simplification, _and_ network serialisation. This test could therefore presumably fail if either the serialisation code was broken, or there was something funky going on in the simplification that meant it couldn't be serialised properly. We want the number of things that can cause our tests to fail to be as small as possible. Is there a way to reduce the number of things that can make this test fail?

mfitz (changes requested):

tests/test_core_network.py
```diff
@@ -475,23 +475,39 @@ def test_attempt_to_simplify_already_simplified_network_throws_error():
...
+
 
+def test_simplifing_puma_network_results_in_correct_record_of_removed_links_and_expected_graph_data(
+        puma_network_pre_simplify, puma_network_post_simplify):
+    n = puma_network_pre_simplify
+
+    link_ids_pre_simplify = set(dict(n.links()).keys())
+
+    deleted_self_loops, n = puma_network_post_simplify
     assert n.is_simplified()
```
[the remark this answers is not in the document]
KasiaKoz: So it used to be done in every test function that checked for different behaviour of simplified network. I moved it into a fixture to save some time, it wasn't huge but gave back a few minutes iirc.

genet/core.py
```diff
@@ -210,13 +210,37 @@ def initiate_crs_transformer(self, epsg):
         else:
             self.transformer = None
 
-    def simplify(self, no_processes=1):
+    def simplify(self, no_processes=1, keep_loops=False):
+        """
+        Simplifies network graph, retaining only nodes that are junctions
+        :param no_processes: Number of processes to split some computation across. The method is pretty fast though
+            and 1 process is often preferable --- there is overhead for splitting and joining the data.
+        :param keep_loops: bool, simplification often leads to
```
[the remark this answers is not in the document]
mfitz: BUT WHAT DOES SIMPLIFICATION LEAD TO?!! The suspense is killing me!

genet/core.py
```diff
@@ -210,13 +210,37 @@ def initiate_crs_transformer(self, epsg):
...
+                         'to PT stops.')
+            if not keep_loops:
+                logging.info('These loops will now be removed. To disable this behaviour, use `keep_loops=True`.'
+                             'Investigate the change log for more information about these links')
+                self.remove_links(to_remove)
+
         # mark graph as having been simplified
         self.graph.graph["simplified"] = True
 
+        return to_remove
```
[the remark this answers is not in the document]
KasiaKoz: ah, maybe I should rename this, I want to return the same IDs either way, so if `keep_loops` is true someone knows easily what was removed, can also go to the changelog for that but i feel this is cleaner, and with the changelog there could be other remove events

tests/test_core_network.py
KasiaKoz: nah, so these are only a small subset of specific stops that were affected by the bug, the majority of other PT stops were fine. I guess there is a way to pull out all of those stops with some conditions, but that will require using non trivial amount of genet code which we're not testing here, so I'm not sure which is better ?

tests/test_core_network.py
KasiaKoz: I have to test saving a simplified network because the data types change (because we merge multiple links, they will have an average or max of a numeric attribute but for example osm tags will all be retained in a set or list and I need to be sure that that network is still valid under matsim. I would like to keep using a puma network that we simplify here because it's close to what we deal with in real life. I added something like this, is that a good compromise? If something goes wrong with simplification, it's caught and a message directs away from this test? 

Screenshot 2022-02-10 at 12 11 11

.github/workflows/build_pipeline.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         pip install -e .
     - name: Lint with flake8
       run: |
-        flake8 . --max-line-length 120 --count  --show-source --statistics --exclude=scripts,tests,notebooks
+        flake8 . --max-line-length 120 --count  --show-source --statistics --exclude=scripts,tests,notebooks,venv
```
mfitz, line 35: Lately I've been thinking that we should be linting our tests too.

tests/test_core_network.py
```diff
@@ -475,38 +475,63 @@ def test_attempt_to_simplify_already_simplified_network_throws_error():
...
+    for stop in puma_network.schedule.stops():
+        assert puma_network.link(stop.linkRefId)['from'] == puma_network.link(stop.linkRefId)['to']
+        assert puma_network.link(stop.linkRefId)['length'] == 1
+
+
+def test_simplified_network_saves_to_correct_dtds(tmpdir, network_dtd, schedule_dtd, puma_network):
+    try:
+        puma_network.simplify()
+    except Exception as e:
+        raise RuntimeError(f"Error simplifying network: {e}, check other simplification tests")
```
mfitz, line 533: I wasn't thinking about `simplify` possibly throwing an exception - if it does that, it will be clear what caused the test to fail, without wrapping and rethrowing the exception. I don't think rethrowing the exception adds anything.

What I was driving at is that we're calling `simplify` - what if it doesn't throw an exception, **but** it's broken in some way? Testing around that and that alone would have a network, simplify it, then make assertions about it (without serialising it before making those assertions). I think we have tests like that, right?

But if we also write the network and it doesn't validate against the DTD, is that because `simplify` has a bug, or because `write_to_matsim` has a bug? Or both? In an ideal world, we test `write` by saying "given a known network, when we serialise it to disk, it should blah, blah (validate against the DTD, etc.)". Whether or not the network was simplified first shouldn't matter, we're just testing that a network serialises properly.

It seems like the thing we're testing here is that a simplified network serialises correctly, so is there a way to have a simplified network with which to test that **without** calling simplify? If that's possible, it would be a better way to test the serialisation of the simplified network. As it stands, this is kind of a pipeline of two operations with some assertions at the end of the pipeline.

genet/core.py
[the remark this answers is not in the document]
KasiaKoz: I decided against this in the end, no one has any business with link IDs that don't exist. After removal, those IDs can be reused for other things so this would just get confusing --- something I didn't think of before and came bit me in the ass in NZ the other day.

tests/test_core_network.py
KasiaKoz: moved to a fixture with a descriptive name

tests/test_core_network.py
KasiaKoz: changed to testing a dedicated fixture with data schema representative of a simplified network

tests/test_core_network.py
```diff
@@ -475,38 +475,63 @@ def test_attempt_to_simplify_already_simplified_network_throws_error():
...
-    assert (set(n.link_id_mapping.keys()) & new_links) == new_links
+    # check the links removed in simplification are all mapped to new links
+    assert set(puma_network.link_simplification_map.keys()) == deleted_links
+    # check even the removed loop links have the simplification mapping retained
+    assert set(puma_network.link_simplification_map.values()) == new_links | deleted_self_loops
+    # check all new links int he network have edge mappings
+    assert (set(puma_network.link_id_mapping.keys()) & new_links) == new_links
 
-    report = n.generate_validation_report()
+    report = puma_network.generate_validation_report()
```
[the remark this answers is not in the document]
KasiaKoz: moved to two separate tests

tests/test_core_network.py
```diff
@@ -475,51 +475,116 @@ def test_attempt_to_simplify_already_simplified_network_throws_error():
...
+                         'length': 52.765151087870265,
+                         'attributes': {'osm:way:highway': {'name': 'osm:way:highway',
+                                                            'class': 'java.lang.String',
+                                                            'text': {'unclassified', 'other'}}}})
+    return n
 
-    assert n.is_simplified()
 
-    link_ids_post_simplify = set(dict(n.links()).keys())
+def test_simplifing_puma_network_results_in_correct_record_of_simplified_links(puma_network):
```
mfitz, line 528: I like this - easier for my brain to parse :-)

tests/test_core_network.py
```diff
@@ -475,51 +475,116 @@ def test_attempt_to_simplify_already_simplified_network_throws_error():
...
+
+def test_simplify_keeps_pt_stop_loops(puma_network):
+    puma_network.simplify()
+    for stop in puma_network.schedule.stops():
+        assert puma_network.link(stop.linkRefId)['from'] == puma_network.link(stop.linkRefId)['to']
+        assert puma_network.link(stop.linkRefId)['length'] == 1
+
+
+def test_simplified_network_saves_to_correct_dtds(tmpdir, network_dtd, network_with_simplified_schema):
+    network_with_simplified_schema.write_to_matsim(tmpdir)
```
mfitz, line 580: 👍

mfitz (approved):

PR #110, opened 2022-03-29
Generalise road pricing type

KasiaKoz:
Gets rid of hard coded Cordon type road pricing. Allows user to specify the toll type, descriptions and allows for change in filenames. Defaults to link type tolling to align with hermes matsim module.

### DIFF
--- a/genet/__init__.py
+++ b/genet/__init__.py
@@ -1,6 +1,6 @@
 from genet.core import Network  # noqa: F401
 from genet.schedule_elements import Schedule, Service, Route, Stop  # noqa: F401
-from genet.use.road_pricing import Cordon  # noqa: F401
+from genet.use.road_pricing import Toll  # noqa: F401
 from genet.inputs_handler.read import *  # noqa: F401,F403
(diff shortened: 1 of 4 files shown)


akay-arup (approved):
@KasiaKoz this is great -- you can feel free to kill my existing branch. Great minds think alike, I guess!

Currently, we output our rp schemes in the following fashion:
```xml
<link id = "1">
    <time = "time1" amount = "amt1">
    <time = "time2" amount = "amt2>
</link>
<link id="2">
    <time = "time1" amount = "amt1">
    <time = "time2" amount = "amt2>
</link>
```
I think this covers us any which way, but in future it might be good to have the option to produce less verbose files for simpler schemes, which I [think are allowable](https://www.matsim.org/apidocs/core/0.4.0/org/matsim/roadpricing/package-summary.html):
```xml
<links>
    <link id="1" />
    <link id="2" />
<links>
<time = "time1" amount = "amt1">
<time = "time2" amount = "amt2">
```

akay-arup:
(the above xml isn't precisely as required, but i think gets the point across)

PR #111, opened 2022-03-29
Upgrade nbconvert dependency version

mfitz:
The [notebooks smoke tests were failing in GitHub Actions builds](https://github.com/arup-group/genet/runs/5734236896?check_suite_focus=true) due to an error from `jinja2`, which is a transitive dependency of the `nbconvert` module we're using. 

Moving to the latest `nbconvert` [fixes this](https://github.com/arup-group/genet/runs/5735692884?check_suite_focus=true):

Screen Shot 2022-03-29 at 12 21 57

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -34,7 +34,7 @@ matplotlib==3.2.1
 memory-profiler==0.57.0
 more-itertools==8.3.0
 munch==2.5.0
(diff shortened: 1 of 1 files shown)


KasiaKoz (approved):
awesomesauce

PR #112, opened 2022-04-01
Add/Remove Services and Routes in groups

KasiaKoz:
Before each add/remove of a Service or Route object to/from Schedule was processed individually. When modifying PT Schedules on a larger scale this takes a long time to process. This change is not some huge smart re-write of existing methods, rather grabbing some low hanging fruit in a fe loops that led to some major improvements. The biggest slow-down was (I think) due to updating the change log individually---these changes now take advantage of the big batch logging for changelog.

Some test for Services on a small test network (adding 10 services each with 3 routes inside)
Before:
Screenshot 2022-04-01 at 17 00 49
After:
Screenshot 2022-04-01 at 17 00 59
Removing services, Before:
Screenshot 2022-04-01 at 17 01 06
After:
Screenshot 2022-04-01 at 17 01 13

I attempted this to speed things up in NZ. Adding ~300 services used to take 40mins, now takes ~2mins 💪


tests/test_core_schedule.py
```diff
@@ -851,6 +881,22 @@ def test_removing_service_updates_vehicles(schedule):
...
+    assert set(s.route_ids()) == {'2', '1', '4', '3'}
+    s.remove_services(['service1', 'service2'])
+    assert set(s.service_ids()) == set()
+    assert set(s.route_ids()) == set()
+
+
+def test_removing_multiple_services_updates_changelog(schedule_graph):
+    s = Schedule(_graph=schedule_graph)
+    s.remove_services(['service1', 'service2'])
+    list(s.change_log().iloc[-2:][['change_event', 'old_id']].to_records()) == [(0, 'remove', 'service1'), (1, 'remove', 'service2')]
```
mfitz, line 896: Send out the assertion search squad! 😄 

I had a mess about with this test, or at least the other one like it, and when I turned this statement into an assertion it failed. I think it was to do with the types of the objects in the two lists we're comparing; one was holding tuples and the other was holding numpy record objects, so they _looked_ the same, but failed equality comparison. I'm guessing there is a quick and easy "pandonic" way to make sure our expectation list has records, not tuples (or the actual list has tuples rather than records)?

tests/test_core_schedule.py
```diff
@@ -851,6 +881,22 @@ def test_removing_service_updates_vehicles(schedule):
...
+def test_multiple_services_are_no_longer_present_in_schedule_after_removing(schedule_graph):
+    s = Schedule(_graph=schedule_graph)
+    assert set(s.service_ids()) == {'service1', 'service2'}
+    assert set(s.route_ids()) == {'2', '1', '4', '3'}
+    s.remove_services(['service1', 'service2'])
+    assert set(s.service_ids()) == set()
+    assert set(s.route_ids()) == set()
+
+
+def test_removing_multiple_services_updates_changelog(schedule_graph):
```
mfitz, line 893: I like these test names 👍

tests/test_core_schedule.py
```diff
@@ -877,7 +923,43 @@ def test_adding_route_with_clashing_id(schedule, route):
...
+    assert set(schedule.service_ids()) == {'service'}
+    assert set(schedule.route_ids()) == {'2', '1'}
+    schedule.add_routes({'service': routes_to_add})
+    assert set(schedule.service_ids()) == {'service'}
+    assert set(schedule.route_ids()) == {'2', 'route_to_add2', '1', 'route_to_add1'}
+
+
+def test_adding_multiple_routes_updates_changelog(schedule, services_to_add):
+    schedule.add_services(services_to_add)
+    list(schedule.change_log().iloc[-2:][['change_event', 'new_id']].to_records()) == [(0, 'add', 'route_to_add1'), (1, 'add', 'route_to_add2')]
```
mfitz, line 962: Another missing assert, I think?

mfitz (changes requested):
I love this performance win!

There are a couple of tests that need some minor tweaking (they currently cannot fail no matter what), so I'm not approving, but they will be quick fixes 👍

tests/test_core_schedule.py
```diff
@@ -763,6 +760,39 @@ def test_adding_service(schedule, service):
...
+
+def test_multiple_services_are_present_in_schedule_after_adding(schedule, services_to_add):
+    schedule.add_services(services_to_add)
+    assert set(schedule.service_ids()) == {'new_service_1', 'new_service_0', 'service'}
+    assert set(schedule.route_ids()) == {'new_route_0_0', '1', '2', 'new_route_1_1', 'new_route_1_0', 'new_route_0_1'}
+
+
+def test_adding_multiple_services_updates_changelog(schedule, services_to_add):
+    schedule.add_services(services_to_add)
+    list(schedule.change_log().iloc[-2:][['change_event', 'new_id']].to_records()) == [(0, 'add', 'new_service_0'), (1, 'add', 'new_service_1')]
```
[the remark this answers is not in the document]
KasiaKoz: 🤦‍♀️

genet/utils/persistence.py
```diff
@@ -11,6 +12,24 @@ def ensure_dir(direc):
             logging.warning(e)
 
 
+def setify(value: Union[str, list, set]):
```
mfitz, line 15: These don't seem like persistence-related helper functions - maybe we should move them later on.

tests/test_core_schedule.py
```diff
@@ -783,14 +783,21 @@ def services_to_add():
 
 
 def test_multiple_services_are_present_in_schedule_after_adding(schedule, services_to_add):
+    existing_routes = set(schedule.route_ids())
```
mfitz, line 786: This test is much nicer now 😄

tests/test_core_schedule.py
```diff
@@ -783,14 +783,21 @@ def services_to_add():
...
-    assert set(schedule.route_ids()) == {'new_route_0_0', '1', '2', 'new_route_1_1', 'new_route_1_0', 'new_route_0_1'}
+
+    assert set(schedule.service_ids()) == expected_services_after_adding
+    assert set(schedule.route_ids()) == expected_routes_after_adding
 
 
 def test_adding_multiple_services_updates_changelog(schedule, services_to_add):
     schedule.add_services(services_to_add)
-    list(schedule.change_log().iloc[-2:][['change_event', 'new_id']].to_records()) == [(0, 'add', 'new_service_0'), (1, 'add', 'new_service_1')]
+    assert list(schedule.change_log().iloc[-len(services_to_add):][['change_event', 'new_id']].itertuples(
```
mfitz, line 799: Now with added asserts!

tests/test_core_schedule.py
```diff
@@ -883,18 +890,17 @@ def test_removing_service_updates_vehicles(schedule):
...
 
-def test_removing_multiple_services_updates_changelog(schedule_graph):
+def test_removing_multiple_services_updates_changelog(schedule, schedule_graph):
     s = Schedule(_graph=schedule_graph)
-    s.remove_services(['service1', 'service2'])
-    list(s.change_log().iloc[-2:][['change_event', 'old_id']].to_records()) == [(0, 'remove', 'service1'), (1, 'remove', 'service2')]
-
+    services_to_remove = list(s.service_ids())
+    s.remove_services(services_to_remove)
+    assert list(s.change_log().iloc[-len(services_to_remove):][['change_event', 'old_id']].itertuples(
```
mfitz, line 902: 👍

tests/test_core_schedule.py
```diff
@@ -1203,8 +1214,9 @@ def test_service_is_no_longer_present_after_removing_all_its_routes(schedule_gra
 
 def test_removing_multiple_routes_updates_changelog(schedule_graph):
     s = Schedule(_graph=schedule_graph)
-    s.remove_routes(['1', '2'])
-    list(s.change_log().iloc[-2:][['change_event', 'old_id']].to_records()) == [(0, 'remove', '1'), (1, 'remove', '2')]
+    route_to_remove=['1', '2']
+    s.remove_routes(route_to_remove)
+    assert list(s.change_log().iloc[-len(route_to_remove):][['change_event', 'old_id']].itertuples(index=False,name=None)) == [('remove', r) for r in route_to_remove]
```
mfitz, line 1219: 👍

tests/test_utils_persistence.py
```diff
@@ -33,6 +33,32 @@ def test_swallows_exceptions_making_new_directories(mocker):
...
 
 
+def test_setifying_list_returns_expected_set():
+    assert persistence.setify(['ABC', 2]) == {'ABC', 2}
+
+def test_setifying_string_returns_expected_single_item_set():
+    assert persistence.setify('ABC') == {'ABC'}
+
+def test_setifying_set_makes_no_difference():
+    assert persistence.setify({'ABC', 2}) == {'ABC', 2}
```
mfitz, line 43: Do we expect the original set, or a semantically equivalent copy of that set?

mfitz (approved):
Smashing!

genet/utils/persistence.py
KasiaKoz: yeah I have a really hard time naming things 🥲  what name would you give them?

tests/test_utils_persistence.py
KasiaKoz: yeah it won't be the original set, but I only care about the contents 😄

ana-kop:
40 minutes to 2 minutes is massive, well done @KasiaKoz !!

PR #113, opened 2022-04-06
Reduce Docker image size

mfitz:
## Before

```bash
REPOSITORY               TAG        IMAGE ID       CREATED             SIZE
genet-local-master       latest     3e95afde36e9   30 seconds ago      1.75GB
```

## After
```bash
REPOSITORY            TAG        IMAGE ID       CREATED             SIZE
genet-local           latest     61c9395a63bf   26 minutes ago      1.46GB
```
Still pretty chonky, even after reclaiming 300 megs. Questions @KasiaKoz:
- Do we still need that Node installation? I think it was for something to do with Kepler, right?
- `notebooks/7. Visualising Network.ipynb` is over 40 MB in size - is that expected? Seems outrageously big for a notebook.

## Validating the image
I connected to the container and ran both the unit tests and the notebook smoke tests. The smoke tests all passed, but one unit test failed. However, when I ran the unit tests in the container created from the image in `master`, I saw the same failure, so this does not seem to have been caused by my changes.

Screen Shot 2022-04-06 at 02 23 04
Screen Shot 2022-04-06 at 02 16 25

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+**/__pycache__
+**/.pytest_cache/
+.git
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,6 @@ RUN curl -sL https://deb.nodesource.com/setup_17.x | bash -
 RUN apt-get -y install nodejs && node --version && npm --version &&  \
 rm -rf /var/lib/apt/lists/*
 
-#RUN apt-get install -y coinor-cbc &&  \
-#rm -rf /var/lib/apt/lists/*
-
 RUN /usr/local/bin/python -m pip install --no-cache-dir --compile --upgrade pip
 
 COPY ./scripts .
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,20 +2,22 @@ FROM python:3.7-slim-stretch
 
 RUN apt-get update && \
 apt-get upgrade -y && \
-apt-get -y install gcc git libspatialindex-dev curl &&  \
+apt-get -y install gcc git libspatialindex-dev curl coinor-cbc &&  \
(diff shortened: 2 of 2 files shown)


KasiaKoz (approved):
Awesome, thanks for de-chonkifying genet, doing Gods work here @mfitz 

We do still need node, kepler is still a functionality.. and you can generate and save those plots to html so it's possible someone would use that via docker.

Now onto the failing test in docker. Wawaweewa 👀 It looks like the snap and route functionality / maybe the max stable set optimisation gives different results inside---but why and what are the consequences 👀  a bit worried tbh. Should be investigated and I feel like we should have some automatic testing within docker... we do use genet within docker frequently 😰

mfitz:
> Awesome, thanks for de-chonkifying genet, doing Gods work here @mfitz
> 
> We do still need node, kepler is still a functionality.. and you can generate and save those plots to html so it's possible someone would use that via docker.
> 
> Now onto the failing test in docker. Wawaweewa 👀 It looks like the snap and route functionality / maybe the max stable set optimisation gives different results inside---but why and what are the consequences 👀 a bit worried tbh. Should be investigated and I feel like we should have some automatic testing within docker... we do use genet within docker frequently 😰

We can add testing inside the container to the CodeBuild pipeline, we have a mechanism to do that by just adding a `run-tests.sh` to the root of the project - see [here](https://github.com/arup-group/ade-tf-modules/blob/2a4fee02a6d4e4260879eb350e1eefbcac113c95/v0.12/codebuild/files/buildspec_docker.yml#L33). This will fail the build right now though, so we should figure out what it going wrong with that test first.

PR #114, opened 2022-04-08
Loosen version constraints for various pytest-related dependencies

mfitz:
When trying to install `scop`:
Screen Shot 2022-04-08 at 13 27 46

This is a spurious version clash - it is almost certain that both apps/libs will work with either version of `pytest-cov`.

We could decide to loosen the constraints for more than just these libraries, of course, but for now these will hopefully at least fix Scop's problem. Opinions welcome.

I may create a similar PR for MC that is much more flexible on the test dependencies, but still specific about non-test dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,6 @@ decorator==4.4.2
 descartes==1.1.0
 dictdiffer==0.8.1
 docutils==0.15.2
-Fiona==1.8.20
 flake8>=3.8.4
 future==0.18.2
 geopandas==0.9.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,6 +15,7 @@ decorator==4.4.2
 descartes==1.1.0
 dictdiffer==0.8.1
 docutils==0.15.2
+Fiona==1.8.20
 flake8>=3.8.4
 future==0.18.2
 geopandas==0.9.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -15,6 +15,7 @@ decorator==4.4.2
 descartes==1.1.0
 dictdiffer==0.8.1
 docutils==0.15.2
+Fiona==1.8.20
```
mfitz, line 18: This is really likely to be something specific to my own machine, but I get fatal errors out of Fiona unless I pin to this specific version. I'm planning on getting a new machine soon, so hopefully this is just temporary, and I'm happy to remove this and keep the pinning as a local thing on my machine only if anybody objects to it being here.

KasiaKoz: i have a vague memory that we deleted this from requirements as it was failing with windows users; geopandas handles fiona install, maybe if we're adding this, we should put it after geopandas on the list? (I think it installs top to bottom... it wont be alphabetical anymore though 😱 )

mfitz: If it's tricky, and it only really help me, I think maybe I will just remove it and manage that locally on my machine. I low key hate these libraries with native dependencies 😆

KasiaKoz (approved):

PR #120, opened 2022-06-14
LAB-1717: Add functionality for adding slope attribute to links

ana-kop:
GeNet currently has functionality for adding elevation values as an attribute to nodes. This PR creates functionality to find elevation values for each end node of a link and then calculate the link slope and add it as an attribute.

### DIFF
--- a/genet/core.py
+++ b/genet/core.py
@@ -2229,7 +2229,7 @@ class Network:
                 link_slope = 0
             else:
                 # calculate slope by dividing the difference between elevations of two nodes by distance between them
-                link_slope = abs(z_1 - z_2) / length
+                link_slope = z_2 - z_1 / length
 
             slope_dict[link_id] = {'slope': link_slope}
 
--- a/genet/core.py
+++ b/genet/core.py
@@ -2203,7 +2203,8 @@ class Network:
             elevation_dict[node_id] = {'z': z}
 
         if run_validation is True:
-            print(elevation.validation_report_for_node_elevation(elevation_dict))
+            report = elevation.validation_report_for_node_elevation(elevation_dict)
+            logging.info(report['summary'])
 
         return elevation_dict
 
--- a/genet/core.py
+++ b/genet/core.py
@@ -2210,11 +2210,11 @@ class Network:
 
     def get_link_slope_dictionary(self, elevation_dict):
         """
-        Takes an elevation raster file in .tif format, finds the elevation of the end nodes for each link, calculates
-        link slope and adds returns a dictionary of link IDs and their slopes; can then use
+        Takes a dictionary of z-value for each network node (as created by get_node_elevation_dictionary() function,
+        calculates link slope and returns a dictionary of link IDs and their slopes; can then use
         self.apply_attributes_to_links() function to add slope as a link attribute to the network.
-        :param elevation_dict: contains node_id as key and elevation in meters as value
-        :return: dict in format {link_id : {'slope': slope}}
+        :param elevation_dict: dict in format {node_id : {'z': z}}
+        :return: dict in format {link_id : {'slope': slope}}, where slope is a float
         """
         slope_dict = {}
 
@@ -2225,14 +2225,14 @@ class Network:
             z_1 = elevation_dict[node_1]['z']
             z_2 = elevation_dict[node_2]['z']
 
-            # TO-DO: calculate crow-fly distance between the 2 nodes
+            # TO-DO: calculate crow-fly distance between the 2 nodes instead of using routed distance
             length = self.link(link_id)['length']
 
             if length == 0:
                 link_slope = 0
             else:
                 # calculate slope by dividing the difference between elevations of two nodes by distance between them
-                link_slope = z_2 - z_1 / length
+                link_slope = (z_2 - z_1) / length
 
             slope_dict[link_id] = {'slope': link_slope}
 
--- a/genet/core.py
+++ b/genet/core.py
@@ -2248,6 +2248,37 @@ class Network:
             raise MissingElevationException('Network nodes do not contain elevation data. '
                                             'Cannot generate validation report.')
 
+    def add_slope_attribute_to_links(self, elevation_tif_file_path, null_value: float):
+        """
+        Takes an elevation raster file in .tif format, finds the elevation of the end nodes for each link, calculates
+        link slope and adds it as an attribute to links.
+        :param elevation_tif_file_path: path to the elevation raster file in .tif format
+        :param null_value: value that represents null in the elevation raster file
+        :return:
+        """
+        img = elevation.get_elevation_image(elevation_tif_file_path)
+        slope_dict = {}
+
+        for link in self.links():
+            link_id = link[0]
+            node_1 = self.link(link_id)['from']
+            node_2 = self.link(link_id)['to']
+            link_length = self.link(link_id)['length']
+
+            z_1 = elevation.get_elevation_data(img, lat=self.node(node_1)['lat'], lon=self.node(node_1)['lon'])
+            z_2 = elevation.get_elevation_data(img, lat=self.node(node_2)['lat'], lon=self.node(node_2)['lon'])
+
+            for z in [z_1, z_2]:
+                # zero values handling - may wish to add infilling based on nearby values later
+                if z == null_value:
+                    z = 0
+
+            # To calculate slope divide the difference between the elevations of two nodes by the distance between them
+            link_slope = abs(z_1 - z_2)/link_length
+            slope_dict[link_id] = {'slope': link_slope}
+
+        self.apply_attributes_to_links(slope_dict)
+
 
 class MissingElevationException(Exception):
     pass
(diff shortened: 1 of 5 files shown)


genet/core.py
```diff
@@ -2248,6 +2248,37 @@ def validation_report_for_node_elevation(self, low_limit=-50, mont_blanc_height=
...
+            z_2 = elevation.get_elevation_data(img, lat=self.node(node_2)['lat'], lon=self.node(node_2)['lon'])
+
+            for z in [z_1, z_2]:
+                # zero values handling - may wish to add infilling based on nearby values later
+                if z == null_value:
+                    z = 0
+
+            # To calculate slope divide the difference between the elevations of two nodes by the distance between them
+            link_slope = abs(z_1 - z_2)/link_length
+            slope_dict[link_id] = {'slope': link_slope}
```
mfitz, line 2278: We're using the absolute value of the difference between the source and target node elevations, so whether we're going uphill or downhill, the value will always be positive, right? This makes me think it will always appear as if we're sloping uphill. Is that correct? I can't tell if that's a problem or not because I don't know how these slope values will be subsequently used - can you flesh that out for me?

Also, we need some unit tests around this function.

mfitz (changes requested):
We need some unit tests around this new function. [The build is broken](https://github.com/arup-group/genet/actions/runs/2495310302) because the code coverage has now dropped below the allowable threshold:

Screen Shot 2022-06-14 at 15 29 38

Let me know if you need some help writing the tests.

genet/core.py
mfitz: >  Downhill is more desirable to cyclists for example than uphill 😄

Unless the cyclist is @gac55 , in which case all bets are off 🤣

ana-kop: That is a very good point @mfitz thank you! I missed that when writing the first version! Fixed it now

genet/core.py
```diff
@@ -760,6 +761,7 @@ def add_link(self, link_id: Union[str, int], u: Union[str, int], v: Union[str, i
             raise RuntimeError('Multi index key needs to be an integer')
 
         self.link_id_mapping[link_id] = {'from': u, 'to': v, 'multi_edge_idx': multi_edge_idx}
+        # TODO calculate length of link if not provided
```
KasiaKoz, line 764: https://github.com/arup-group/genet/issues/122

genet/core.py
```diff
@@ -2200,54 +2202,37 @@ def add_elevation_to_nodes(self, elevation_tif_file_path, null_value: float):
                 z = 0
             elevation_dict[node_id] = {'z': z}
 
-        self.apply_attributes_to_nodes(elevation_dict)
+        if run_validation is True:
+            print(elevation.validation_report_for_node_elevation(elevation_dict))
```
KasiaKoz, line 2206: Not sure about printing this dictionary. In some (bad) cases I guess this dict can get pretty big (the 'too high' and 'too low' values are reported for each node). 
Instead of print, let's use `logging.info` and report only the summary part of the dict? and if that looks bad, someone can run the validation and get the output as a dict that they can do something with.

scripts/add_link_slope_to_network.py
```diff
@@ -35,12 +34,18 @@
                             help='Output directory for the updated network',
                             required=True)
 
+    arg_parser.add_argument('-sj',
+                            '--save_jsons',
```
KasiaKoz, line 38: I'd call this `save_validation_jsons` or just `save_validation`; network and schedule can be saved to json to so that will make the param less vague

scripts/add_link_slope_to_network.py
```diff
@@ -52,15 +57,34 @@
         epsg=projection
     )
 
-    logging.info('Adding elevation to network nodes')
-    n.add_elevation_to_nodes(elevation_tif_file_path=elevation, null_value=tif_null_value)
+    logging.info('Creating elevation dictionary for network nodes')
+    elevation_dictionary = n.get_node_elevation_dictionary(elevation_tif_file_path=elevation, null_value=tif_null_value)
+
+    with open(os.path.join(output_dir, 'elevation/node_elevation_dictionary.json'), 'w', encoding='utf-8') as f:
```
KasiaKoz, line 63: this has two problems:

1. we're hardcoding this slash `/` which means a windows user will have problems (more than they usually do anyway); you can do this: `os.path.join(output_dir, 'elevation', 'node_elevation_dictionary.json')` --- the method can join multiple strings into a path

2. more important: we're only calling ensure dir on the `output_dir`, not `output_dir/elevation` so that folder will not exist and the save will fail. I suggest doing:
```
elev_out_dir = os.path.join(output_dir, 'elevation')
ensure_dir(elev_out_dir)
```
and use `elev_out_dir` later on

KasiaKoz (changes requested):
Some changes needed to make paths to save work but otherwise nice work!

We moved from a script for just elevation to just slope. Maybe it would be nice to have a script with flags where we can specify we want: both, or one or the other ?

genet/core.py
ana-kop: Yeah that sounds like a good idea!

scripts/add_link_slope_to_network.py
ana-kop: good point!

genet/core.py
```diff
@@ -2200,54 +2202,38 @@ def add_elevation_to_nodes(self, elevation_tif_file_path, null_value: float):
...
 
-            report = {
-                'summary': {'total_nodes': len(elevation_list),
-                            'min_value': int(min_value),
-                            'max_value': int(max_value),
-                            'mean': int(mean),
-                            'median': int(median),
-                            'extremely_high_values_count': len(too_high),
-                            'extremely_low_values_count': len(too_low)},
+            # TO-DO: calculate crow-fly distance between the 2 nodes
```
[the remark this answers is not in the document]
ana-kop: Yes we are using routed distance now to make it simpler, because calculating crow-fly distance requires us to import another module to get the formula... @KasiaKoz said it would be quite bulky, so we decided to use this as proxy for now

scripts/add_link_slope_to_network.py
```diff
@@ -0,0 +1,98 @@
...
+
+    arg_parser.add_argument('-el',
+                            '--elevation',
+                            help='Path to the elevation tif file',
+                            required=True)
+
+    arg_parser.add_argument('-nv',
+                            '--null_value',
+                            help='Value that represents null in the elevation tif file',
+                            required=True)
```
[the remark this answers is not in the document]
ana-kop: No it is usually a random negative number, defined for the file in the source. E.g. for the test data that I sourced it was -32768

tests/test_core_network.py
```diff
@@ -2977,41 +2977,33 @@ def test_saving_network_to_csv(network1, correct_schedule, tmpdir):
...
+    return elevation_dictionary
 
 
-def test_validation_report_for_network_without_node_elevation(network3):
-    n = network3
-    with pytest.raises(MissingElevationException) as error_info:
-        n.validation_report_for_node_elevation()
-    assert "do not contain elevation" in str(error_info.value)
+def test_getting_link_slope_dictionary(network3, elevation_dictionary):
+    slope_dict = network3.get_link_slope_dictionary(elevation_dictionary)
```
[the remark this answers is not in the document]
ana-kop: Yeah I felt like I would just be rewriting the function itself, but let's do that

mfitz (approved):

genet/core.py
```diff
@@ -2200,54 +2202,38 @@ def add_elevation_to_nodes(self, elevation_tif_file_path, null_value: float):
...
 
-            for node_id, node_elev in nodes_dictionary.items():
-                if node_elev < low_limit:
-                    too_low[node_id] = node_elev
-                elif node_elev > mont_blanc_height:
-                    too_high[node_id] = node_elev
+        for link in self.links():
+            link_id = link[0]
+            node_1 = self.link(link_id)['from']
+            node_2 = self.link(link_id)['to']
```
KasiaKoz, line 2224: since you're using the `self.links()` iterator anyway, you can write it more simply this way:
```
for link_id, link_data in self.links():
        node_1 = link_data['from']
        node_2 = link_data['to']
        ...
        length = link_data['length']
```

genet/utils/elevation.py
```diff
@@ -13,3 +14,48 @@ def get_elevation_data(img, lat, lon):
...
     return elevation_meters
+
+
+def validation_report_for_node_elevation(elev_dict, low_limit=-50, mont_blanc_height=4809):
+    """
+    Generates a validation report for the node elevation dictionary.
+    :param elev_dict: contains node_id as key and elevation in meters as value
+    :param low_limit: values below this param get flagged as possibly wrong; set at -50m (below sea level) by default,
+    can optionally set a different value
+    :param mont_blanc_height: values above this param get flagged as possibly wrong; defaults to 4809m,
```
KasiaKoz, line 25: lol, so how did we arrive at mont blanc as a benchmark? how come not mont everest? 😂

KasiaKoz (approved):
cool beans, just had one comment on using the `links` iterator

PR #124, opened 2022-06-23
Magical java typing for additional attributes

KasiaKoz:
Mentioned in 'Future Work' of https://github.com/arup-group/genet/pull/118 and addition to that branch of work. This PR is a change in addition to https://github.com/arup-group/genet/pull/118 and allows the representation of additional attributes in a new, simpler format. (please note the branch of a branch merging request 😬)

Old, long form format:
```python
'attributes': {
        'additional_attrib': {'name': 'additional_attrib', 'class': 'java.lang.String', 'text': 'attrib_value'}
    }
```
is still allowed, and encouraged when the user wants to be specific about java types of the values.

This PR makes GeNet use simple form attributes:
```python
'attributes': {
        'additional_attrib': 'attrib_value' 
    }
```
across the board, i.e. when
- reading xml files,
- setting 'simplified' tag for the network
- crs for network and schedule
- reading osm (i.e., when creating network graph from osm, the format of osm tags being saved)
- road pricing osm ID to toll matching

## Breaking changes, in some cases

Because genet now represents all additional attributes in simple form, the following possible problems will need to be announced (via new release of genet with changelog).

- In project scripts that search for links with values of specific osm tag values. Example:
```
links = n.extract_links_on_edge_attributes(
    conditions= {'attributes': {'osm:way:highway': {'text': 'primary'}}},
)
```
this will break because genet now represents this data in short form, do the way to query it is:
```
links = n.extract_links_on_edge_attributes(
    conditions= {'attributes': {'osm:way:highway': 'primary'}},
)
```

## Other annoying things

The road pricing functionality in genet that allows OSM ID to toll mappings is affected by this change. The problem is that now the data can be many types. In particular, OSM IDs can be integers (genet's type when reading OSM), float (PT2MATSim network's type via being saved as `java.lang.Long`) and strings (puma's default when creating network from OSM). This is dealt with via an additional parameter in the offending function, the user needs to be mindful of their data types and set them accordingly to match. This was also noted in the update of the example jupyter notebook for road pricing. All defaults are set to strings which should mean no breaking changes in anyone's current setup.

Please note, this PR includes a complete rerun of many jupyter notebooks and updates to some documentation stored whithin, which I think will make it look a lot bigger than it actually is...


tests/test_data/matsim/network.xml
```diff
@@ -15,7 +15,7 @@
 			<attributes>
 				<attribute name="osm:way:access" class="java.lang.String" >permissive</attribute>
 				<attribute name="osm:way:highway" class="java.lang.String" >unclassified</attribute>
-				<attribute name="osm:way:id" class="java.lang.Long" >26997928</attribute>
+				<attribute name="osm:way:id" class="java.lang.Float" >26997928.0</attribute>
```
mfitz, line 18: There is some potential inconsistency in the "round trip", where we read in a network with some attributes typed as `java.lang.Long`, they become Python floats, and then if we write them out again they will be, I think, given a `class` of `java.lang.Float`. Do I have that right?

I am struggling to reason about this - could that change cause any problems? I guess that both `java.lang.Long` and `java.lang.Float` become floats in Python anyway, so as far as GeNet is concerned, there is no difference, but could MATSim's behaviour be effected somehow by the difference in type?

tests/test_output_matsim_xml_writer.py
```diff
@@ -226,38 +229,55 @@ def test_saving_network_with_geometry_produces_correct_polyline_in_link_attribut
...
+        tmpdir, network_with_badly_formatted_attributes_and_geometry):
+    network_with_badly_formatted_attributes_and_geometry['network'].write_to_matsim(tmpdir)
 
     found_geometry_attrib = False
     for event, elem in ET.iterparse(os.path.join(tmpdir, 'network.xml'), events=('start', 'end')):
         if event == 'start':
             if elem.tag == 'attribute':
                 if elem.attrib['name'] == 'geometry':
-                    assert elem.text == '_ibE_seK_ibE_ibE_ibE_ibE'
+                    assert elem.text == network_with_badly_formatted_attributes_and_geometry['encoded_geometry']
```
mfitz, line 280: 👍

mfitz (changes requested):
I've left a handful of minor comments.

I've also left a comment about how a network could be changed purely by the round trip of reading in to GeNet and then writing out without making any modifications (attributes with an original `class=java.lang.Long` will become `class=java.lang.Float`), that has raised a flag in my brain. I can't think through the ramifications of this behaviour, but it seems worth trying to get a handle on whether or not this change is important, particularly in terms of the effect on MATSim.

genet/input/matsim_reader.py
```diff
@@ -115,15 +116,19 @@ def update_additional_attrib(elem, attribs):
...
 def read_additional_attrib(elem):
-    d = elem.attrib
     if elem.text is None:
-        d['text'] = ''
-        logging.warning(f"Elem {elem.attrib['name']} is being read as None.")
+        t = ''
+        logging.warning(f"Elem {elem.attrib['name']} is being read as None. Defaulting value to empty string.")
     elif (',' in elem.text) and elem.attrib['name'] != 'geometry':
-        d['text'] = set(elem.text.split(','))
+        t = set(elem.text.split(','))
```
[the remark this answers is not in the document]
KasiaKoz: yeah, this only really applies to modes on the network, in which case yeah it's fine. I'm thinking what additional attributes would need duplicates and also preserve order order but i'm coming up with nothing. I logged it as an issue here: https://github.com/arup-group/genet/issues/125 and https://arupdigital.atlassian.net/browse/LAB-1750

tests/test_java_dtypes.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+from genet.utils.java_dtypes import *
+
+
+def test_mapping_java_array_to_python_type():
```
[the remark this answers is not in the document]
KasiaKoz: indeed, will add

KasiaKoz:
@mfitz the two biggest things that I tried to address post your review:
- reading comma separated values to sets, did not fix right away as I don't think it's going to cause massive issues at this point in time but logged it: https://github.com/arup-group/genet/issues/125 and https://arupdigital.atlassian.net/browse/LAB-1750
- the Java-python, python-Java functions are not bijective, even more so now that I added the extra java classes. To this effect, I added a flag to force reading matsim networks using the old, long form in case the java types need to be preserved. This also has an added bonus of backwards compatibility which will be appreciated by owners of current project scripts.

genet/output/matsim_xml_writer.py
```diff
@@ -35,21 +43,47 @@ def retain_allowed_attributes_for_xml(d, elem_type):
...
             attribs_to_delete = []
             for attrib, value in attribs['attributes'].items():
-                try:
-                    attribs['attributes'][attrib]['name']
-                    attribs['attributes'][attrib]['class']
-                    attribs['attributes'][attrib]['text']
-                except Exception as e:
-                    logging.warning(f'Attempt to access required keys in data under "attributes:{attrib}" key '
-                                    f'resulted in {type(e)} with message "{e}".')
+                if not (is_of_matsim_format(value) or can_be_put_in_matsim_format(value)):
```
mfitz, line 80: Ah, sweet, sweet boolean logic! 😄

genet/utils/java_dtypes.py
```diff
@@ -0,0 +1,37 @@
+JAVA_DTYPE_MAP = {
+    'java.lang.Array': list,
+    'java.lang.Boolean': bool,
+    'java.lang.Double': float,
+    'java.lang.Float': float,
+    'java.lang.Long': float,
+    'java.lang.Integer': int,
+    'java.lang.Byte': int,
```
mfitz, line 8: 👍

tests/test_java_dtypes.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_mapping_java_bool_to_python_type():
+    assert java_to_python_dtype('java.lang.Boolean') == bool
+
+
+def test_mapping_java_double_to_python_type():
+    assert java_to_python_dtype('java.lang.Double') == float
+
+
+def test_mapping_java_long_to_python_type():
```
mfitz, line 17: 👍

genet/output/matsim_xml_writer.py
```diff
@@ -35,21 +43,47 @@ def retain_allowed_attributes_for_xml(d, elem_type):
...
 
+def is_of_matsim_format(attribute_value):
+    if isinstance(attribute_value, dict):
+        if {'name', 'class', 'text'}.issubset(set(attribute_value.keys())):
+            return True
+    return False
+
+
+def can_be_put_in_matsim_format(attrib_value):
+    return type(attrib_value) in java_dtypes.PYTHON_DTYPE_MAP
```
mfitz, line 54: 👍

mfitz (approved):
Looks good!

PR #126, opened 2022-07-06
Issue #117: links_on_spatial_condition failing for large spatial area

### LINKED ISSUE #117: links_on_spatial_condition failing for large spatial area
Hey, 

I was using the function `links_on_spatial_condition` with a geojson input and it was failing due to a memory error (see below). The geojson contained a polygon which covered multiple cities. I was able to work around it by using `gpd.sjoin(network_gdf,urban_gdf,how='inner', op='intersects')`. It might be worth changing the function `links_on_spatial_condition` to use `gpd.sjoin` instead of `gdf.intersects`. 

Thanks. 

Error:
'Segmentation fault (core dumped)'

ana-kop:
Addresses issue #117 :

Using the function links_on_spatial_condition with a geojson input results in an error due to a memory error:
```
Segmentation fault (core dumped)
```

The geojson contained a polygon which covered multiple cities. Possible to work around it by using `gpd.sjoin(network_gdf,urban_gdf,how='inner', op='intersects')`. It might be worth changing the function links_on_spatial_condition to use `gpd.sjoin` instead of `gdf.intersects`.

### DIFF
--- a/genet/core.py
+++ b/genet/core.py
@@ -600,13 +600,20 @@ class Network:
         self.remove_mode_from_links(diff_links, mode)
 
     def _find_ids_on_geojson(self, gdf, how, geojson_input):
-        shapely_input = spatial.read_geojson_to_shapely(geojson_input)
-        return self._find_ids_on_shapely_geometry(gdf=gdf, how=how, shapely_input=shapely_input)
+        if how == 'intersect':
+            geojson_input_gdf = gpd.read_file(geojson_input)
+            return list(gpd.sjoin(gdf, geojson_input_gdf, how='inner', op='intersects')['id'])
+        elif how == 'within':
+            shapely_input = spatial.read_geojson_to_shapely(geojson_input)
+            return list(gdf[gdf.within(shapely_input)]['id'])
+        else:
+            raise NotImplementedError('Only `intersect` and `contain` options for `how` param.')
 
     def _find_ids_on_shapely_geometry(self, gdf, how, shapely_input):
         if how == 'intersect':
-            return list(gdf[gdf.intersects(shapely_input)]['id'])
-        if how == 'within':
+            geojson_input_gdf = gpd.GeoDataFrame(index=[0], crs=gdf.crs, geometry=[shapely_input])
+            return list(gpd.sjoin(gdf, geojson_input_gdf, how='inner', op='intersects')['id'])
+        elif how == 'within':
             return list(gdf[gdf.within(shapely_input)]['id'])
         else:
             raise NotImplementedError('Only `intersect` and `contain` options for `how` param.')


genet/core.py
```diff
@@ -600,13 +600,20 @@ def retain_n_connected_subgraphs(self, n: int, mode: str):
...
+            return list(gpd.sjoin(gdf, geojson_input_gdf, how='inner', op='intersects')['id'])
+        elif how == 'within':
+            shapely_input = spatial.read_geojson_to_shapely(geojson_input)
+            return list(gdf[gdf.within(shapely_input)]['id'])
+        else:
+            raise NotImplementedError('Only `intersect` and `contain` options for `how` param.')
 
     def _find_ids_on_shapely_geometry(self, gdf, how, shapely_input):
         if how == 'intersect':
-            return list(gdf[gdf.intersects(shapely_input)]['id'])
```
KasiaKoz, line 608: as far as I understand that's the only line you need to change, the rest of the code falls in place

genet/core.py
```diff
@@ -600,13 +600,20 @@ def retain_n_connected_subgraphs(self, n: int, mode: str):
...
-        if how == 'within':
+            geojson_input_gdf = gpd.GeoDataFrame(index=[0], crs=gdf.crs, geometry=[shapely_input])
+            return list(gpd.sjoin(gdf, geojson_input_gdf, how='inner', op='intersects')['id'])
+        elif how == 'within':
             return list(gdf[gdf.within(shapely_input)]['id'])
```
KasiaKoz, line 617: I assume this is at risk too and should be updated similarly to intersection?

genet/core.py
```diff
@@ -600,13 +600,20 @@ def retain_n_connected_subgraphs(self, n: int, mode: str):
...
+            return list(gdf[gdf.within(shapely_input)]['id'])
+        else:
+            raise NotImplementedError('Only `intersect` and `contain` options for `how` param.')
 
     def _find_ids_on_shapely_geometry(self, gdf, how, shapely_input):
         if how == 'intersect':
-            return list(gdf[gdf.intersects(shapely_input)]['id'])
-        if how == 'within':
+            geojson_input_gdf = gpd.GeoDataFrame(index=[0], crs=gdf.crs, geometry=[shapely_input])
+            return list(gpd.sjoin(gdf, geojson_input_gdf, how='inner', op='intersects')['id'])
```
KasiaKoz, line 615: `op` is deprecated, `predicate` should be used instead: https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin.html

KasiaKoz (changes requested):
A lot of code duplication happening here, I don't think you need to change anything in https://github.com/arup-group/genet/blob/4fcc3d36a49f8b042e7987082802ee61f3c2fc34/genet/core.py#L632 

---
Now that we're not relying on these `intersects` and `within` methods, it would be cleaner and more general to  get rid of the if statements and open up the `predicate` param: https://geopandas.org/en/stable/docs/reference/api/geopandas.sindex.SpatialIndex.valid_query_predicates.html so the user can choose from the geopandas list of operations and our code would simplify to:
```python
def _find_ids_on_shapely_geometry(self, gdf, predicate, shapely_input):
    geojson_input_gdf = ...
    return list(gpd.sjoin(gdf, geojson_input_gdf, how='inner', predicate=predicate)['id'])
```

mfitz:
Closing as will not fix.

PR #130, opened 2022-07-14
Improve notebooks smoke test report

mfitz:
#### Running locally
Screen Shot 2022-07-14 at 20 19 03

#### Running in GitHub Actions
Screen Shot 2022-07-14 at 19 49 44

#### Running locally
Screen Shot 2022-07-14 at 20 12 35

#### Running in GitHub Actions
Screen Shot 2022-07-14 at 20 04 18

## Additional Functionality
You can now optionally pass one or more notebook paths as params to the python script, where before you could pass only a directory from which the list of notebooks would be discovered.

This was mostly just to give me a faster feedback loop than running all the notebooks while working on these changes, but could still be generally useful in situations where you're making changes to one or two notebooks and want the same sort of fast feedback. For example:

```bash
python notebooks/notebook_smoke_tests.py -k genet \
-n 'notebooks/1. Intro.ipynb' \
-n 'notebooks/2.4. Reading CSV data.ipynb' \
-n 'notebooks/3.3. Writing GTFS data.ipynb' \
-n 'notebooks/4.3. Using Network - Routing.ipynb' \
-n 'notebooks/4.4. Using Network - Auxiliary Files referencing Network IDs.ipynb'
```

Screen Shot 2022-07-14 at 19 56 58
Screen Shot 2022-07-14 at 21 17 27

**NB:** The notebooks (and unit tests) that rely on the solver fail on my machine because I don't have the underlying native solver installed. I get `ApplicationError: No executable found for solver 'cbc'`. I don't consider it a priority to address that. It turned out that those failures were quite handy for me while I was making these changes, because I wanted to see a mixture of passing and failing notebooks in the summary.

### DIFF
--- a/notebooks/notebook_smoke_tests.py
+++ b/notebooks/notebook_smoke_tests.py
@@ -66,13 +66,13 @@ def install_jupyter_kernel(kernel_name):
 
 def execute_notebook(notebook_path):
     print("Executing notebook '{}{}{}'...".format(Fore.YELLOW, notebook_path, Style.RESET_ALL))
-    kernel_install_cmd = [
+    cmd = [
         'jupyter', 'nbconvert',
         '--to', 'notebook',
         '--execute', '"{}"'.format(notebook_path),
         '--output-dir=/tmp'
     ]
-    return run_shell_command(kernel_install_cmd)
+    return run_shell_command(cmd)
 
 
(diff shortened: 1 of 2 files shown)


KasiaKoz (approved):
Nice one

PR #134, opened 2022-08-05
Fixing the add_link_slope_to_network.py script

ana-kop:
I found a couple of small bug in the add_link_slope_to_network.py script while using it for the NZ test network. This PR fixes them, and also adds an option to save the node elevation as an attribute onto the network (not just link slope as before)


scripts/add_link_slope_to_network.py
```diff
@@ -3,9 +3,11 @@
...
 
+import genet
 from genet import read_matsim
 from genet.utils.persistence import ensure_dir
-import genet.outputs_handler.sanitiser as sanitiser
+import genet.output.sanitiser as sanitiser
+import genet.utils.elevation as elevation
 
 if __name__ == '__main__':
     arg_parser = argparse.ArgumentParser(description='Add elevation data to network nodes and validate it')
```
KasiaKoz, line 13: add something about computing link slopes to description and that you can do both or just one of them

KasiaKoz (changes requested):
Looks good, but I have some thoughts :D 
I think we should add an argument `write_slope_to_network`, default to `True` and then we have covered all options, slope, z values for nodes, one of them or both, it'll be nice to have this flexibility I think.
I think I liked the old name for the script `add_elevation_to_network.py` more, the link slope one is a bit of a mouthful and slope is a consequence of elevation so maybe good to have it vague

scripts/add_elevation_to_network.py
```diff
@@ -35,6 +37,16 @@
...
 
+    arg_parser.add_argument('-we',
+                            '--write_elevation_to_network',
+                            help='Whether node elevation data should be written as attribute to the network; defaults to True',
+                            default=True)
+
+    arg_parser.add_argument('-ws',
+                            '--write_slope_to_network',
+                            help='Whether link slope data should be written as attribute to the network; defaults to True',
+                            default=True)
```
KasiaKoz, line 48: let's add `type=bool` here

scripts/add_elevation_to_network.py
```diff
@@ -35,6 +37,16 @@
                             help='Output directory for the updated network',
                             required=True)
 
+    arg_parser.add_argument('-we',
+                            '--write_elevation_to_network',
+                            help='Whether node elevation data should be written as attribute to the network; defaults to True',
+                            default=True)
```
KasiaKoz, line 43: let's add `type=bool` here

scripts/add_elevation_to_network.py
```diff
@@ -35,6 +37,16 @@
...
+
+    arg_parser.add_argument('-ws',
+                            '--write_slope_to_network',
+                            help='Whether link slope data should be written as attribute to the network; defaults to True',
+                            default=True)
+
     arg_parser.add_argument('-sj',
                             '--save_jsons',
                             help='Whether elevation and slope dictionaries and report are saved; defaults to True',
```
KasiaKoz, line 52: let's add `type=bool` here, also description says 'defaults to True' but it doesn't

scripts/add_elevation_to_network.py
```diff
@@ -63,36 +77,50 @@
...
     if save_dict_to_json is True:
-        with open(os.path.join(elevation_output_dir, 'elevation/link_slope_dictionary.json'), 'w',
+        with open(os.path.join(elevation_output_dir, 'link_slope_dictionary.json'), 'w',
                   encoding='utf-8') as f:
-            json.dump(sanitiser.sanitise_dictionary(report), f, ensure_ascii=False, indent=4)
+            json.dump(sanitiser.sanitise_dictionary(slope_dictionary), f, ensure_ascii=False, indent=4)
+
 
-    logging.info('Adding link slope as attribute to the network')
+    if write_slope_to_network is True:
```
KasiaKoz, line 114: you can just do: `if write_slope_to_network` since `write_slope_to_network` is a bool

KasiaKoz (changes requested):
When you do this, you also produce some visual outputs or geojsons? Would be good to include this in this script so these are ready for inspection automatically.

Otherwise some minor comments, one that definitely needs the change (docs don't match with behaviour)

scripts/add_elevation_to_network.py
ana-kop: it has `default=True` on the next line!

KasiaKoz: hah! that bit was collapsed, completely missed it, sorry! :D

KasiaKoz (approved):

Issue #135, opened 2022-08-08
Summary Stats 


KasiaKoz:
Methods to print to console and save to file (via standard outputs) summary statistics about the entire network and schedule.

Let's discuss what statistics should be included.

I envision a method `summary` for both network and schedule; something like this within network:

```python
def summary(self):
   module.summarise_network(self)
   if self.schedule:
       self.schedule.summary()
```

KasiaKoz:
network and schedule have a print methods that show number of links, nodes and services and stops respectively:

```
Graph info: Name: Network graph
Type: MultiDiGraph
Number of nodes: 387573
Number of edges: 883961
Average in degree:   2.2808
Average out degree:   2.2808 
Schedule info: Schedule:
Number of services: 1013
Number of routes: 1398
Number of stops: 14142
```

might be good to report a split by mode perhaps, e.g. `n` bus services (stops dont really have modes, but you can link them up with the routes that use them and those have modes (confer recent multimodal access/egress work)

KasiaKoz:
a simple summary of all modes on the links, and all PT modes

KasiaKoz:
`X.link/node/stop/route/service_attribute_summary(data=True)` gives the data schema, all attributes and additional (nested) attributes for link/node/stop/route/service objects belonging to network/schedule, and an example of 5 values under that attribute, but the underlying method that generates this can be configured differently --- it shows unique values, in some cases there are many, e.g. osm IDs so we limited them to 5, but in some cases, like modes, it would make sense to show all.

KasiaKoz:
stats about headways might be useful (if they are not being saved with standard outputs already) 
https://github.com/arup-group/genet/blob/main/genet/schedule_elements.py#L1526

fredshone:
two cases come up recently:
1) people asking for summary stats of network and schedules, often split by mode. I think well covered above.
2) me getting caught out by the precise naming of stop attributes as required for the intermodal access/egress stuff

pam and mc have summary cli commands which stdout out everything that might be useful, the idea being that someone will read and copy/paste only what they need. So the more the merrier i guess - provided it is clearly presented

ana-kop:
@KasiaKoz @fredshone I have created a method that I think includes most of the things you mention - here is what it gives when used on the TE network + schedule:
```
{'network_graph_info': {'Number of network links': 447610,
  'Number of network nodes': 985170},
 'modes': {'Modes on network links': {'artificial',
   'bike',
   'bus',
   'car',
   'ferry',
[...]

PR #136, opened 2022-08-17
Summary Stats

ana-kop:
Issue: https://github.com/arup-group/genet/issues/135

Creates a method that generates a summary report with descriptive statistics for the network and schedule. A report for the TE network looks as follows:

```
{'network': {'network_graph_info': {'Number of network links': 447611,
   'Number of network nodes': 985204},
  'modes': {'Modes on network links': {'artificial',
    'bike',
    'bus',
    'car',
    'ferry',
    'gondola',
    'rail',
    'stopFacilityLink',
    'subway',
    'tram',
    'walk'},
   'Number of links by mode': {'bike': 813980,
    'artificial': 17415,
    'gondola': 17415,
    'subway': 17464,
    'bus': 171976,
    'stopFacilityLink': 17415,
    'rail': 17969,
    'tram': 17419,
    'ferry': 17427,
    'car': 964771,
    'walk': 858490}},
  'osm_highway_tags': {'Number of links by tag': {'primary': 38304,
    'tertiary': 97991,
    'trunk_link': 7865,
    'bridleway': 54,
    'construction': 2,
    'secondary_link': 75,
    'motorway_link': 2123,
    'proposed': 2,
    'motorway': 2085,
    'footway': 224,
    'service': 350477,
    'tertiary_link': 93,
    'trunk': 66798,
    'secondary': 36269,
    'living_street': 3410,
    'primary_link': 819,
    'steps': 2,
    'track': 90,
    'unclassified': 83345,
    'pedestrian': 34,
    'residential': 288954,
    'cycleway': 1577,
    'path': 34}}},
 'schedule': {'schedule_info': {'Number of services': 1351,
   'Number of routes': 4687,
   'Number of stops': 20198},
  'modes': {'Modes in schedule': {'bus', 'ferry', 'rail', 'subway', 'tram'},
   'Services by mode': {'ferry': 4,
    'subway': 7,
    'bus': 1165,
    'rail': 174,
    'tram': 1},
   'PT stops by mode': {'ferry': 10,
    'subway': 22,
    'bus': 19993,
    'rail': 171,
    'tram': 3}},
  'accessibility_tags': {'Stops with tag bikeAccessible': 202,
   'Unique values for bikeAccessible tag': {'true'},
   'Stops with tag carAccessible': 202,
   'Unique values for carAccessible tag': {'true'}}}}
  ```

### CODE REVIEW

File: genet/core.py [hunk 1]
```diff
@@ -2263,3 +2263,58 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+        logging.info('Creating a summary report')
+        report = {}
+
+        network_stats = {'Number of network links': nx.number_of_nodes(self.graph),
+                         'Number of network nodes': nx.number_of_edges(self.graph)}
+        report['network_graph_info'] = network_stats
+        report['modes'] = {}
+        report['modes']['Modes on network links'] = self.modes()
+
+        highway_tags = self.link_attribute_data_under_key({'attributes': 'osm:way:highway'})
```
KasiaKoz, line 2280: need to add something for 'backwards' compatibility here. It is still possible to read additional attributes in long form, in which case the query would be: `{'attributes': {'osm:way:highway': 'text'}}`. I think there is a simple way to do this, just by checking if `highway_tags` is empty, add a warning and attempt to search using long form --- but you will need to do a test to see if that's right

genet/core.py
```diff
@@ -2263,3 +2263,58 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+            stops_by_modes = {}
+            for mode in schedule_modes:
+                stops_by_modes[mode] = len(self.schedule.stops_on_modal_condition(mode))
+            report['modes']['PT stops by mode'] = stops_by_modes
+
+            report['accessibility_tags'] = {}
+            report['accessibility_tags']['Stops with tag bikeAccessible'] = len(
+                self.schedule.stop_attribute_data({'attributes': 'bikeAccessible'}))
+            report['accessibility_tags']['Stops with tag carAccessible'] = len(
+                self.schedule.stop_attribute_data({'attributes': 'carAccessible'}))
```
KasiaKoz, line 2318: this needs to be a method for the `Schedule` object

KasiaKoz, line 2318: here I would use that pandas object a bit more, dataframe or series, I don't remember anymore, apart from just the length, report also the unique values for those tags, which is something that Fred had issues with in the matsim config: whether the values are 'true' or 'True' etc.

genet/output/geojson.py
```diff
@@ -209,3 +209,8 @@ def generate_standard_outputs(n, output_dir, gtfs_day='19700101', include_shp_fi
             output_dir=os.path.join(output_dir, 'routing'),
             include_shp_files=include_shp_files
         )
+
+    # import json
+    # summary_report = n.summary()
+    # with open(os.path.join(output_dir, 'summary_report.json'), 'w', encoding='utf-8') as f:
+    #     json.dump(sanitiser.sanitise_dictionary(summary_report), f, ensure_ascii=False, indent=4)
```
KasiaKoz, line 216: delete?

tests/test_core_network.py
```diff
@@ -3017,3 +3042,65 @@ def test_getting_link_slope_dictionary(network3):
     link_slope = (z_2 - z_1) / length
     slope_dict = network3.get_link_slope_dictionary(elevation_dictionary)
     assert slope_dict['0']['slope'] == link_slope
+
+
+def test_generating_summary_report(network5):
+    network5.schedule = Schedule(epsg='epsg:27700', services=[
```
KasiaKoz, line 3048: just write that schedule in the fixture, atm that fixture uses `correct_schedule` and then overwrites it in the test

tests/test_core_network.py
```diff
@@ -219,6 +220,20 @@ def network4():
     return n4
 
 
+@pytest.fixture()
+def network5(correct_schedule):
```
KasiaKoz, line 224: so I know you're following 'convention' of networkInteger - but it's really not a great convention, let's be better than that 😂 i would call this fixture: `network_for_summary_stats` and also include the expected output in the fixture, so it would return a dictionary: {'network_object': n, 'summary_output': ...} and then your test at the bottom will be ~ three lines

KasiaKoz (changes requested):
looking good, just needs some shuffling around

genet/output/geojson.py
ana-kop: I actually needed to keep this, but I needed to fix the test with the old way of presenting attributes first (`{'attributes': {'osm:way:highway': 'text'}}`) - I have addressed the earlier comment now so it works with both notations now

genet/core.py
ana-kop: Yep that's a really good point, I have added a check for that so that both cases can be handled

genet/core.py
ana-kop: I moved this whole thing to the geojson module, so it is a function now, rather than a Network or Schedule method

genet/output/geojson.py
```diff
@@ -150,6 +153,83 @@ def generate_standard_outputs_for_schedule(schedule, output_dir, gtfs_day='19700
         os.path.join(output_dir, 'trips_per_day_per_route_aggregated_per_stop_name_pair.csv'))
 
 
+def summary(network):
```
KasiaKoz, line 156: do separate methods for network and schedule objects

KasiaKoz (changes requested):
So you moved the summary methods to the geojson module, but you're not doing anything with geodataframes or geojsons here, so I think you should put them somewhere else. Crucially though, there need to be separate methods for the network and schedule, and the network method should call on the schedule method if there is pt. The smaller the functions are the easier they are to test and maintain. 

I would do this:
- `summarise` or `summarise_network` method in `genet.use.network` (doesn't exist you'd need to make that module)
- `summarise` or `summarise_schedule` method in `genet.use.schedule`

And then also add class methods `summarise` for both network and schedule that call on those methods (I think you had a class method for Network to begin with?). 

That would be my preference I think, but equally you could just move that code straight to the class method. All I care about is that there are class methods to summarise the two objects, and a schedule can be summarised independently of a network object.

KasiaKoz (approved):
Can you also add a cell showcasing this new summary functionality to this example notebook: (there is a summary section at the start) https://github.com/arup-group/genet/blob/main/notebooks/4.1.%20Using%20Network:%20Accessing%20Data.ipynb

fredshone (approved):
sure - slight concern that the stop tags eg `bikeAccessible` are hardcoded, but i see it fits well with existing patterns.

PR #140, opened 2022-08-24
Add method for splitting link at given point

ana-kop:
Network method `split_link_at_point()` takes a link and point coordinates, and splits the link at the point to create 2 new links; the old link is then deleted.

Parameters:
`link_id`: ID of the link to split
`point_y`: y-coordinates of the point to split at
`point_x`: x-coordinates of the point to split at


genet/core.py
```diff
@@ -2263,3 +2264,48 @@ def get_link_slope_dictionary(self, elevation_dict):
             slope_dict[link_id] = {'slope': link_slope}
 
         return slope_dict
+
+    def split_link_at_point(self, link_id, point_lat, point_lon, new_node_id):
```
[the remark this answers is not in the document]
KasiaKoz: probably safer to remove the user defined node and use https://github.com/arup-group/genet/blob/4fcc3d36a49f8b042e7987082802ee61f3c2fc34/genet/core.py#L1919 to find a unique node ID that doesn't already exist in the network. Otherwise you'll have to warn the user of changing node IDs

genet/core.py
```diff
@@ -2263,3 +2264,48 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+        to_node = self.link(link_id)['to']
+        from_node = self.node(from_node)
+        to_node = self.node(to_node)
+        line = LineString([(float(from_node['lon']), float(from_node['lat'])),
+                           (float(to_node['lon']), float(to_node['lat']))])
+
+        # if not, find nearest point on the link line
+        # (not using 'contains' method due to too high accuracy required to evaluate to True)
+        if line.distance(point) > 1e-8:
+            point = line.interpolate(line.project(point))
```
KasiaKoz, line 2291: this point is not being used after this, no?
I think we should add a step where (if there is one) the complex geometry is split at that point and given to the new links

genet/core.py
```diff
@@ -2263,3 +2264,48 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+
+        self.add_node(new_node_id, {'id': new_node_id, 'lon': point.x, 'lat': point.y})
+
+        # create 2 new links: from_node -> new_node ; new_node -> to_node
+        new_link_1 = link_id + '_1'
+        new_link_2 = link_id + '_2'
```
KasiaKoz, line 2297: not a huge fan of these, especially if the same link gets divided multiple times, it'll end up looking really crazy, use https://github.com/arup-group/genet/blob/4fcc3d36a49f8b042e7987082802ee61f3c2fc34/genet/core.py#L1945 to generate unique, unsused IDs

genet/core.py
```diff
@@ -2263,3 +2264,48 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+        self.add_link(new_link_1, from_node['id'], new_node_id)
+        self.add_link(new_link_2, new_node_id, to_node['id'])
+
+        # apply attributes from the old link to the 2 new links
+        old_link_attributes = self.link(link_id)
+        for k in ['id', 'from', 'to', 'length', 's2_from', 's2_to']:
+            old_link_attributes.pop(k, None)
+
+        self.apply_attributes_to_link(new_link_1, old_link_attributes)
+        self.apply_attributes_to_link(new_link_2, old_link_attributes)
```
KasiaKoz, line 2308: Consolidate the attributes and add links with correct attributes to begin with

KasiaKoz (changes requested):

genet/core.py
ana-kop: this is the point at which the link gets split in the next line of code:
```
self.add_node(new_node_id, {'id': new_node_id, 'x': point.x, 'y': point.y})
```

 so this step updates the point coords if the point given by user does not lie on the link

genet/core.py
```diff
@@ -2263,3 +2264,44 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+        # if not, find nearest point on the link line
+        # (not using 'contains' method due to too high accuracy required to evaluate to True)
+        if line.distance(point) > 1e-8:
+            point = line.interpolate(line.project(point))
+
+        self.add_node(new_node_id, {'id': new_node_id, 'x': point.x, 'y': point.y})
+
+        # create 2 new links: from_node -> new_node ; new_node -> to_node
+        new_link_1 = self.generate_index_for_edge()
+        print(new_link_1)
```
KasiaKoz, line 2296: delete this print

tests/test_core_network.py
```diff
@@ -3017,3 +3017,14 @@ def test_getting_link_slope_dictionary(network3):
...
+
+def test_splitting_link_at_point(network1):
+    network1.split_link_at_point('0', 51.52287873320000, 0.14625948709420000)
+    correct_link_1_attributes = {'from': '101982', 'to': '101987', 'id': '1', 'freespeed': 4.166666666666667,
+                                 'capacity': 600.0, 'permlanes': 1.0, 'oneway': '1', 'modes': ['car'],
+                                 'attributes': {'osm:way:access': {'name': 'osm:way:access', 'class': 'java.lang.String', 'text': 'permissive'},
+                                                'osm:way:highway': {'name': 'osm:way:highway', 'class': 'java.lang.String', 'text': 'unclassified'},
+                                                'osm:way:id': {'name': 'osm:way:id', 'class': 'java.lang.Long', 'text': '26997928'},
+                                                'osm:way:name': {'name': 'osm:way:name', 'class': 'java.lang.String', 'text': 'Brunswick Place'}}}
+    assert network1.link('1') == correct_link_1_attributes
```
KasiaKoz, line 3030: should also add a check for the other link

tests/test_core_network.py
```diff
@@ -3017,3 +3017,14 @@ def test_getting_link_slope_dictionary(network3):
     link_slope = (z_2 - z_1) / length
     slope_dict = network3.get_link_slope_dictionary(elevation_dictionary)
     assert slope_dict['0']['slope'] == link_slope
+
+
+def test_splitting_link_at_point(network1):
+    network1.split_link_at_point('0', 51.52287873320000, 0.14625948709420000)
+    correct_link_1_attributes = {'from': '101982', 'to': '101987', 'id': '1', 'freespeed': 4.166666666666667,
```
KasiaKoz, line 3024: this is a good place to use mocker, that way you will get the ids for the new node and two new links that you expect, but which are not part of the test. So this would look like this:

```python
def test_splitting_link_at_point(network1, mocker):
    new_node_ID = 'new_node_ID'
    new_link_1_ID = 'new_link_1_ID'
    new_link_2_ID = 'new_link_2_ID'
    mocker.patch.object(Network, 'generate_index_for_node', return_value=new_node_ID)
    mocker.patch.object(Network, 'generate_index_for_edge', return_value=new_link_1_ID)
    mocker.patch.object(Network, 'generate_index_for_edge', return_value=new_link_2_ID)
```
and then you can use the `new_node_ID, new_link_1_ID, new_link_2_ID` downstream. **I highly recommend this approach** - because we might change the ID generation methods one day, we've always been on the look out for something more robust and less random for a long time

tests/test_core_network.py
KasiaKoz: also need a check that old link is not in the network anymore

KasiaKoz (changes requested):
Needs a couple of minor changes (rogue print method will be messy in the logging and a couple of checks in the test missing), otherwise great. 
Maybe add a logging info after all is done that links with this node have been added? (since the node can move a little bit?)

genet/core.py
```diff
@@ -2263,3 +2264,48 @@ def get_link_slope_dictionary(self, elevation_dict):
...
+        """
+
+        # check if point is on the link LineString
+        point = Point(point_lon, point_lat)
+        from_node = self.link(link_id)['from']
+        to_node = self.link(link_id)['to']
+        from_node = self.node(from_node)
+        to_node = self.node(to_node)
+        line = LineString([(float(from_node['lon']), float(from_node['lat'])),
+                           (float(to_node['lon']), float(to_node['lat']))])
```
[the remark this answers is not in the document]
KasiaKoz: links with complex geometry (that was generated through simplification) will have a `'geometry'` attribute - I added that in, and your line thing still stands as a fallback ✔️

KasiaKoz (approved):

Issue #82, opened 2021-06-24
Check attributes for negative values


gac55:
I have some MATSim warnings for a network that passed GeNet validation:

```
2021-06-24T18:47:40,300  WARN LinkImpl:120 capacity=-600.0 of link id 7857587829102305401_7857587832100126133 may cause problems
2021-06-24T18:47:40,300  WARN LinkImpl:122  Future occurences of this logging statement are suppressed.
2021-06-24T18:47:40,301  WARN LinkImpl:138 permlanes=-1.0 of link id 7857587829102305401_7857587832100126133 may cause problems
```

Should we check for negative permlanes and capacities?

KasiaKoz:
https://arupdigital.atlassian.net/browse/LAB-1725

Issue #19, opened 2020-08-20
need check for np.inf linkspeeds


fredshone:
`"inf"` link speeds cause matsim to fail, eg:

```<link id="pt_150042002011" from="pt_150042002011" to="pt_150042002011" freespeed="inf" capacity="9999.0" permlanes="1.0" oneway="1" modes="bus,artificial,stopfacilitylink" length="1.0"/>```

KasiaKoz:
https://arupdigital.atlassian.net/browse/LAB-588

PR #141, opened 2022-08-25
Validation zhuzh-up: link attribute value checks

KasiaKoz:
Resolves https://github.com/arup-group/genet/issues/19 - Check for infinite values
Resolves https://github.com/arup-group/genet/issues/82 - Check for negative values

As a bonus this PR also adds checks for fractional values and re-organises the functions that provide the conditions for the checks - it was starting to get very crowded in the validation method. 

Fixed a bug I didn't know existed - the method was not testing nested values for additional attributes, it's good that I caught that now that we're using them more!

It also improves the logging of any issues with the network, e.g.:
```
2022-08-25 16:32:52,084 - Checking validity of the Network
2022-08-25 16:32:52,084 - Checking validity of the Network graph
2022-08-25 16:32:52,084 - Defaulting to checking graph connectivity for modes: ['car', 'walk', 'bike']. You can change this by passing a `modes` param
2022-08-25 16:32:52,084 - Checking network connectivity for mode: car
2022-08-25 16:32:52,084 - The graph for mode: car has: 3 connected components, 1 sinks/dead_ends and 1 sources/unreachable nodes.
2022-08-25 16:32:52,084 - The graph has more than one connected component for mode car! If this is not expected, consider using the `connect_components` method to connect the components, or `retain_n_connected_subgraphs` with `n=1` to extract the largest component
2022-08-25 16:32:52,084 - Checking network connectivity for mode: walk
2022-08-25 16:32:52,084 - The graph for mode: walk has: 0 connected components, 0 sinks/dead_ends and 0 sources/unreachable nodes.
2022-08-25 16:32:52,084 - Checking network connectivity for mode: bike
2022-08-25 16:32:52,084 - The graph for mode: bike has: 0 connected components, 0 sinks/dead_ends and 0 sources/unreachable nodes.
2022-08-25 16:32:52,085 - Checking link values for `length`
2022-08-25 16:32:52,085 - 1 of links have zero values for `length`
2022-08-25 16:32:52,085 - 1 of links have fractional values for `length`
2022-08-25 16:32:52,085 - Checking link values for `capacity`
2022-08-25 16:32:52,085 - 1 of links have negative values for `capacity`
2022-08-25 16:32:52,085 - 1 of links have infinite values for `capacity`
2022-08-25 16:32:52,085 - Checking link values for `freespeed`
2022-08-25 16:32:52,085 - 1 of links have zero values for `freespeed`
2022-08-25 16:32:52,085 - 2 of links have negative values for `freespeed`
2022-08-25 16:32:52,086 - Checking link values for `modes`
2022-08-25 16:32:52,086 - Checking link values for `{'attributes': {'osm:way:id': 'text'}}`
2022-08-25 16:32:52,086 - 1 of links have zero values for `{'attributes': {'osm:way:id': 'text'}}`
```

Compared to:
```
2022-08-25 16:32:15,334 - Checking validity of the Network
2022-08-25 16:32:15,335 - Checking validity of the Network graph
2022-08-25 16:32:15,335 - Checking network connectivity for mode: car
2022-08-25 16:32:15,336 - Checking network connectivity for mode: walk
2022-08-25 16:32:15,336 - Checking network connectivity for mode: bike
2022-08-25 16:32:15,337 - 1 of links have values of 0 for `length`
2022-08-25 16:32:15,337 - 1 of links have values of 0 for `freespeed`
2022-08-25 16:32:15,337 - 1 of links have values of 0 for `id`
```
before.

Nb. More PRs in the validation theme coming up soon. I'm leaving updating the notebook: [6.1. Validating Network: MATSim Specific](https://github.com/arup-group/genet/blob/main/notebooks/6.1.%20Validating%20Network:%20MATSim%20Specific.ipynb) to last in the series.


genet/core.py
```diff
@@ -2066,6 +2067,41 @@ def zero_value(value):
             }
         return report
 
+    def report_on_link_attribute_condition(self, attribute, condition):
```
mfitz, line 2070: I'm liking the smaller functions 👍

genet/core.py
```diff
@@ -2066,6 +2067,41 @@ def zero_value(value):
...
+        links_satifying_condition = self.extract_links_on_edge_attributes(conditions=conditions)
+        return {
+            'number_of': len(links_satifying_condition),
+            'percentage': len(links_satifying_condition) / self.graph.number_of_edges(),
+            'link_ids': links_satifying_condition
+        }
+
+    def check_connectivity_for_mode(self, mode):
+        logging.info(f'Checking network connectivity for mode: {mode}')
+        # subgraph for the mode to be tested
```
mfitz, line 2090: I don't know if these comments are adding anything - they're not really explaining anything the code doesn't already make clear. And if the code _isn't_ making something clear, the best approach is to clarify with better names, etc.

genet/utils/dict_support.py
```diff
@@ -113,6 +113,11 @@ def combine_edge_data_lists(l1, l2):
     return [(u, v, dat) for (u, v), dat in edges.items()]
 
 
+def dict_to_string(d):
+    # turn a (potentially) nested dictionary into a string
```
mfitz, line 117: Again - is this comment adding anything? The comment pretty much echoes the function name.

genet/validate/network.py
```diff
@@ -35,3 +36,33 @@ def describe_graph_connectivity(G):
     # find number of connected subgraphs
     dict_to_return['number_of_connected_subgraphs'] = len(find_connected_subgraphs(G))
     return dict_to_return
+
+
+def zero_value(value):
+    return value in {0, '0', '0.0'}
```
mfitz, line 42: I guess this fails for the string `"0.00"`? If that might turn out to be important, you could parse the string to an int and then check for equality to 0.

genet/validate/network.py
```diff
@@ -35,3 +36,33 @@ def describe_graph_connectivity(G):
...
+
+
+def negative_value(value):
+    if isinstance(value, str):
+        return '-' in value
```
mfitz, line 47: I could pass the string `non-integer` and be told it is negative. Do we want that? Might be safer to parse to int and perform a numerical check. There is also the string function `startswith()`, which seems a better fit than `in` here.

genet/validate/network.py
```diff
@@ -35,3 +36,33 @@ def describe_graph_connectivity(G):
...
+    return math.isinf(value)
+
+
+def fractional_value(value):
+    if isinstance(value, str):
+        return ('0.' in value) and (value != '0.0')
+    return 1 > value > 0
+
+
+LINK_ATTRIBUTE_VALIDATION_TOOLBOX = {
```
mfitz, line 63: This feels a little bit like an impoverished class. What made you go with a dict here?

tests/test_core_network.py
```diff
@@ -2656,8 +2657,12 @@ def test_generate_validation_report_with_pt2matsim_network(network_object_from_t
                 'bike': {'problem_nodes': {'dead_ends': [], 'unreachable_node': []},
                          'number_of_connected_subgraphs': 0}},
             'link_attributes': {
-                'links_over_1km_length': {'number_of': 0, 'percentage': 0.0, 'link_ids': []},
-                'zero_attributes': {}}},
+                'links_over_1000_length': {'number_of': 0, 'percentage': 0.0, 'link_ids': []},
+                'zero_attributes': {},
+                'negative_attributes': {},
+                'infinite_attributes': {},
+                'fractional_attributes': {}
```
mfitz, line 2664: We could be a bit more fine-grained in our tests here. For example, I can imagine a test that uses a network with a zero attribute. We expect that to be reflected in the report. The assertion would be much narrower - grab the `link_attributes` sub dict and assert that it looks how you expect, but ignore the rest of the report. Or go even further and assert only on `report['graph']['link_attributes']['zero_attributes']`.

tests/test_core_network.py
```diff
@@ -2696,8 +2701,12 @@ def test_generate_validation_report_with_correct_schedule(correct_schedule):
                                             'number_of_connected_subgraphs': 0},
                                    'bike': {'problem_nodes': {'dead_ends': [], 'unreachable_node': []},
                                             'number_of_connected_subgraphs': 0}},
-            'link_attributes': {'links_over_1km_length': {'number_of': 0, 'percentage': 0.0, 'link_ids': []},
-                                'zero_attributes': {}}},
+            'link_attributes': {'links_over_1000_length': {'number_of': 0, 'percentage': 0.0, 'link_ids': []},
```
mfitz, line 2704: We've had to change this test, even though its name suggests it is only concerned with the reporting on the schedule, and nothing has changed in that code. This feels like the test telling you that your assertion is too wide. Should this test care at all how the `link_attributes` section of the report looks? That should be tested elsewhere and removed from the equation in this test.

tests/test_core_network.py
```diff
@@ -2717,24 +2726,154 @@ def test_generate_validation_report_with_correct_schedule(correct_schedule):
...
 
+def test_long_links_show_up_in_validation_report():
+    n = Network('epsg:27700')
+    n.add_link('1', 1, 2, attribs={'length': 10000, 'capacity': 1, 'freespeed': 1, "modes": ['car', 'bus']})
+    n.add_link('2', 2, 3, attribs={'length': 2, 'capacity': 1, 'freespeed': 2, "modes": ['car', 'bus']})
+
+    report = n.generate_validation_report()
+
+    assert_semantically_equal(
+        report['graph']['link_attributes']['links_over_1000_length'],
```
mfitz, line 2737: Yes! This is exactly what I'm talking about! The assertion here is narrow enough to protect the test from changes in bits of the code that it does not care about.

tests/test_core_network.py
```diff
@@ -2717,24 +2726,154 @@ def test_generate_validation_report_with_correct_schedule(correct_schedule):
...
+@pytest.mark.parametrize("value,offending_value", offending_link_attribute_values_and_names)
+def test_values_of_ids_are_not_flagged_in_validation_report(value, offending_value):
+    n = Network('epsg:27700')
+    n.add_link(offending_value, 1, 2, attribs={'length': 1, 'capacity': 1, 'freespeed': 1, "modes": ['car', 'bus']})
+    n.add_link('2', 2, 3, attribs={'length': 2, 'capacity': 1, 'freespeed': 2, "modes": ['car', 'bus']})
+
+    report = n.generate_validation_report()
+
+    assert_semantically_equal(
+        report['graph']['link_attributes'][f'{value}_attributes'],
```
mfitz, line 2753: mando-way-mandalorian

tests/test_core_network.py
```diff
@@ -2717,24 +2726,154 @@ def test_generate_validation_report_with_correct_schedule(correct_schedule):
...
+        {
+            'length': {'number_of': 1, 'percentage': 0.5, 'link_ids': ['1']},
+            'capacity': {'number_of': 1, 'percentage': 0.5, 'link_ids': ['1']},
+            'freespeed': {'number_of': 1, 'percentage': 0.5, 'link_ids': ['1']}
+        }
+    )
+
+
+def test_negative_value_attributes_show_up_in_validation_report():
+    n = Network('epsg:27700')
```
mfitz, line 2804: These tests are very good 👏 

The assertions are nice and narrow, **and** I can see clearly, without leaving the test, _why_ we expect what we expect in those assertions

mfitz (changes requested):
I've added a few minor comments - should be small changes if you agree with them.

genet/core.py
```diff
@@ -1999,55 +1999,56 @@ def invalid_network_routes(self):
...
         """
         logging.info('Checking validity of the Network')
         logging.info('Checking validity of the Network graph')
         report = {}
+
         # describe network connectivity
-        modes = ['car', 'walk', 'bike']
-        report['graph'] = {'graph_connectivity': {}}
+        if modes is None:
+            modes = ['car', 'walk', 'bike']
```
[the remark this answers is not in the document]
KasiaKoz: [is dangerooousss](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/#:~:text=In%20Python%2C%20when%20passing%20a,or%20even%20a%20class%20instance.) (pycharm complains)

in this case it would probably be fine, nothing happens to that modes list and I can't really think why one would change that list inside the function... but maybe a bad look ?

genet/validate/network.py
KasiaKoz: mm, good point

genet/validate/network.py
KasiaKoz: ah yes, another good point

genet/validate/network.py
KasiaKoz: lol, I started with a function, but it didnt have any arguments and just returned a dictionary so that's why a dict, might do a quick dataclass here instead

tests/test_core_network.py
KasiaKoz: yeah there are a few of these old blanket tests hanging about, got new and better tests for the link attribute checks, I should just take the opportunity to chop them up and make them into proper tests

genet/core.py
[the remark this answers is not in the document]
mfitz: Ah yes! I forgot about this. I've suffered from that exact bug in the past.

genet/validate/network.py
mfitz: I was thinking more that this dict holds key/value pairs where the value is a validation function, right? So could you have a validation class that has all of those functions as methods? If you need different implementations of those methods, you could extend the class.

genet/core.py
```diff
@@ -2066,6 +2067,41 @@ def zero_value(value):
...
+        :return:
+        """
+        if isinstance(attribute, dict):
+            conditions = dict_support.nest_at_leaf(deepcopy(attribute), condition)
+        else:
+            conditions = {attribute: condition}
+
+        links_satifying_condition = self.extract_links_on_edge_attributes(conditions=conditions)
+        return {
+            'number_of': len(links_satifying_condition),
```
mfitz, line 2083: Can we get the actual broken values into the report as well? Right now I can see that there is a problem with the value for a particular attribute on a given node, but I can't see what the dodgy value was.

KasiaKoz:
Some minor updates plus:

- added dataclasses to represent conditions for testing network link attributes
- added one more condition checking for `None` values

mfitz (approved):

PR #142, opened 2022-08-26
Makes spatial information compulsory when adding new node(s)

ana-kop:


genet/core.py
```diff
@@ -657,23 +657,28 @@ def _find_link_ids_on_s2_geometry(self, gdf, how, s2_input):
...
-        else:
-            # TODO do not add nodes without spatial info
-            self.graph.add_node(node)
+        # check for spatial info
+        if ('lat' not in attribs.keys() and 'lon' not in attribs.keys()) and \
+                ('x' not in attribs.keys() and 'y' not in attribs.keys()):
+            raise RuntimeError('Cannot add Node without spatial information')
+
+        # check for clashing node IDs
+        if node in dict(self.nodes()).keys():
```
KasiaKoz, line 674: there is a dedicated method for this: https://github.com/arup-group/genet/blob/main/genet/core.py#L1836

genet/core.py
```diff
@@ -657,23 +657,28 @@ def _find_link_ids_on_s2_geometry(self, gdf, how, s2_input):
...
         :return:
         """
-        if attribs is not None:
-            self.graph.add_node(node, **attribs)
-        else:
-            # TODO do not add nodes without spatial info
-            self.graph.add_node(node)
+        # check for spatial info
+        if ('lat' not in attribs.keys() and 'lon' not in attribs.keys()) and \
+                ('x' not in attribs.keys() and 'y' not in attribs.keys()):
```
KasiaKoz, line 670: this statement is a bit crazy :D it's difficult to understand with all the ands and nots (btw, to check for a key in dict you dont need the `.keys()`). I think you should write this differently so it's more readable:
```python
keys = set(attribs.keys())
if not ({'lat', 'lon'}.issubset(keys) or {'x', 'y'}.issubset(keys)):
    raise ..

```

genet/core.py
```diff
@@ -686,14 +691,18 @@ def add_nodes(self, nodes_and_attribs: dict, silent: bool = False, ignore_change
...
         :return:
         """
-        # check for clashing nodes
+        # check for spatial info
+        for node_id, value in nodes_and_attribs.items():
+            if ('lat' not in value.keys() and 'lon' not in value.keys()) and \
+                    ('x' not in value.keys() and 'y' not in value.keys()):
+                raise RuntimeError(f'Cannot add Node `{node_id}` without any spatial information')
+
+        # check for clashing node IDs
```
KasiaKoz, line 700: Duplicated code 🙅 `add_node` is nothing special, so should delegate to `add_nodes`. Less code means less code to test and maintain and less code to go wrong.

tests/test_core_network.py
```diff
@@ -853,16 +852,24 @@ def test_link_attribute_data_under_keys_generates_key_for_nested_data(network1):
...
+def test_adding_node_with_clashing_id_throws_error():
     n = Network('epsg:27700')
-    n.add_node(1)
-    assert n.node(1) == {}
-    assert n.graph.has_node(1)
+    n.add_node(1, {'x': 1, 'y': 2, 'a': 1})
+
+    with pytest.raises(RuntimeError) as error_info:
+        n.add_node(1, {'x': 2, 'y': 2, 'a': 2})
+    assert "ID already exists" in str(error_info.value)
```
KasiaKoz, line 872: v nice

KasiaKoz (changes requested):
a few comments there, but the reason I'm sending it back is the code duplication, genet is only growing so we have to be very pedantic about this stuff

genet/core.py
```diff
@@ -704,16 +693,40 @@ def add_nodes(self, nodes_and_attribs: dict, silent: bool = False, ignore_change
...
+                columns=['x', 'y'], index=df_nodes.loc[missing_x_y].index
+            )
+
+        nodes_and_attribs_to_add = dict_support.merge_complex_dictionaries(
+            df_nodes[['x', 'y', 'lon', 'lat', 'id']].T.to_dict(), nodes_and_attribs)
+
+        # pandas is terrible with large numbers so we update them/generate them here
+        for node, attribs in nodes_and_attribs_to_add.items():
+            if 's2_id' not in nodes_and_attribs[node]:
+                attribs['s2_id'] = spatial.generate_index_s2(attribs['lat'], attribs['lon'])
```
KasiaKoz, line 720: this fills in missing spatial info, x,y or lat,lon if only one is passed - which is preffered so people dont have to reproject anything themselves, and s2 ids are added for convenience / math with the rest of the network

ana-kop: some really cool stuff @KasiaKoz ! 🥇

KasiaKoz (approved):

Issue #123, opened 2022-06-17
Validation and methods to deal with isolated nodes


KasiaKoz:
Removing links can lead to lingering isolated nodes. Isolated nodes cause an error in MATSim:

```
java.lang.RuntimeException: For person id=PERSON_ID: getNearestLink returned Null! act=act [type=shop_food][coord=[x=544092.0 | y=713828.0]][linkId=null][startTime=undefined][endTime=15:30:00][duration=undefined][facilityId=null]
```

KasiaKoz:
In the meantime, this be dealt with by using:

```python
import networkx as nx

# n is a genet.Network

unused_nodes = list(nx.isolates(n.graph))
n.remove_nodes(unused_nodes)
```

KasiaKoz:
https://arupdigital.atlassian.net/browse/LAB-1721

PR #143, opened 2022-08-26
Validation zhuzh-up: isolated nodes

KasiaKoz:
Resolves: https://github.com/arup-group/genet/issues/123 by adding methods to:
- check if there are any isolated nodes in the graph
- list their IDs
- remove them

with ample logging. 

Adds info on isolated nodes to validation report:

```python
{'graph': {
...
  'isolated_nodes': {'has_isolated_nodes': 'True', 'nodes': ['1']},
...}
}
```

and adds extra logging while running the report:

```
2022-08-26 11:36:48,203 - Checking validity of the Network
2022-08-26 11:36:48,203 - Checking validity of the Network graph
...
2022-08-26 11:36:48,204 - This Network has isolated nodes! Consider cleaning it up with `remove_isolated_nodes`
...
```

Bonus: corrects misspelt method `remove_unsused_stops` -> `remove_unused_stops` (the PT equivalent of isolated nodes)

Nb. More PRs in the validation theme coming up soon. I'm leaving updating the notebook: [6.1. Validating Network: MATSim Specific](https://github.com/arup-group/genet/blob/main/notebooks/6.1.%20Validating%20Network:%20MATSim%20Specific.ipynb) to last in the series.


genet/core.py
```diff
@@ -1839,6 +1839,20 @@ def has_node(self, node_id):
     def has_nodes(self, node_id: list):
         return all([self.has_node(node_id) for node_id in node_id])
 
+    def has_isolated_nodes(self) -> bool:
+        return bool(self.isolated_nodes())
```
ana-kop, line 1843: Not sure if this is needed as a separate function?

genet/core.py
```diff
@@ -2023,6 +2037,14 @@ def generate_validation_report(self, modes_for_strong_connectivity=None, link_me
             graph_connectivity[mode] = self.check_connectivity_for_mode(mode)
         report['graph'] = {'graph_connectivity': graph_connectivity}
 
+        has_isolated_nodes = self.has_isolated_nodes()
+        report['graph']['isolated_nodes'] = {
+            'has_isolated_nodes': str(has_isolated_nodes),
```
ana-kop, line 2042: I would maybe get rid of this line and then just keep an `isolated_nodes` key for simplicity, so it just shows `None` if there aren't any? Or maybe at least make it a number of isolated nodes, then have a separate list of them under `nodes`?

ana-kop (approved):
Looks great, I just added some comments about whether we really need to have a Boolean which says if there are any isolated nodes, I think maybe a number of isolated nodes would make more sense - but up to you :)

genet/core.py
KasiaKoz: it's just a convenience method, makes the code more readable later on

genet/core.py
KasiaKoz: good idea, i'll make it so:
```python
report['graph']['isolated_nodes'] = {
    "number_of_nodes: 3,
    "nodes" : [A, B, C]
}
```
or 
```python
report['graph']['isolated_nodes'] = {
    "number_of_nodes: 0,
    "nodes" : []
}
```
if there aren't any 👍

Issue #144, opened 2022-08-26
Check PT headway values in validation


KasiaKoz:
Add checks for <s>negative or</s> 0 headway values for PT.

(Negative makes no sense, the order trips does not matter and we can always re-order by time)

KasiaKoz:
https://arupdigital.atlassian.net/browse/LAB-1250

PR #146, opened 2022-09-09
Validation zhuzh-up: headway value checks

KasiaKoz:
Resolves https://github.com/arup-group/genet/issues/144 - Check PT headway values during validation

- Adds what I consider a fix for headway stats computation. We compute headways by looking at the difference in departure time between consecutive trips. In our approach, the value for headway for the first trip is (now) an empty value. Previously it was infilled with zero, which affected the headway stats reported downstream. It's left as empty/`NaN`/`NaT` and is ignored when computing stats later on.
- Exposes the methods to generate trip headways and headway stats `DataFrame`s to `Route` and `Service` objects (originally only `Schedule` could produce these)
- Adds info on headways to validation report and warns of zero headways. Looks like this:
```python
    {'schedule_level': {
        'is_valid_schedule': False, 
        'invalid_stages': ['not_has_valid_services'],
        'has_valid_services': False, 
        'invalid_services': ['service'],
        'headways': {
            'has_zero_min_headways': True,
            'routes': {'number_of_affected': 1, 'ids': ['service_0']},
            'services': {'number_of_affected': 1, 'ids': ['service']}}}, 
     'service_level': ..., 
     'route_level': {'service': {
        'service_0': {
            'is_valid_route': False, 
            'invalid_stages': ['not_has_correctly_ordered_route'],
            'headway_stats': {'mean_headway_mins': 0.0, 'std_headway_mins': 0.0, 'max_headway_mins': 0.0, 'min_headway_mins': 0.0}}}},
     'vehicle_level': ...}}
```

Nb. Part of a wider validation upgrade series. I'm leaving updating the notebook: [6.1. Validating Network: MATSim Specific](https://github.com/arup-group/genet/blob/main/notebooks/6.1.%20Validating%20Network:%20MATSim%20Specific.ipynb) to last in the series.


tests/test_validate_schedule.py
```diff
@@ -37,13 +37,17 @@ def correct_schedule():
...
         'schedule_level': {'is_valid_schedule': True, 'invalid_stages': [], 'has_valid_services': True,
-                           'invalid_services': []},
+                           'invalid_services': [], 'headways': {'has_zero_min_headways': False}},
         'service_level': {
             'service': {'is_valid_service': True, 'invalid_stages': [],
                         'has_valid_routes': True, 'invalid_routes': []}},
         'route_level': {
-            'service': {'1': {'is_valid_route': True, 'invalid_stages': []},
-                        '2': {'is_valid_route': True, 'invalid_stages': []}}},
+            'service': {'1': {'is_valid_route': True, 'invalid_stages': [],
```
elizabethc-arup, line 45: A quick note on an edge case that we ran into in the NZ ABM was with (inter-regional) buses which often only have one service / trip a day. We ended up accounting for them by hardcoding all services with a service/trip length of 1 to a 24-hour headway, not sure how often this would occur however something that we ran into.

elizabethc-arup (approved):
This has been a very helpful function that we use to check whether schedules have been implemented sensibly. A quick note to consider is the edge case that we ran into in the NZ ABM of services with only 1 trip per day, which would result in a 0 or infinite headway rather than 24 hours. :)

tests/test_validate_schedule.py
KasiaKoz: Yeah that's a very good point, in the old version those would default to a 0 headway. I settled on leaving the NaT/NaN because I didn't want to presume.. My thinking is, we could leave the empty for the user to interpret - I think it's not impossible there to be a service that shows up in the schedule but runs once a week and then the headway should ideally be 7*24*60 right? But we can't know about it because we lost the part of GTFS that could tell us about it. All throughout the genet schedule though, we talk about gtfs DAY and the services are assumed to be running in a day, so a 24h default would make sense.

What do you think would make most sense / be the most convenient for a potential user?

Issue #145, opened 2022-09-09
Check for PT speed in validation


KasiaKoz:
Add computation of speeds for PT services and checks for high/low values in Schedule validation

PR #147, opened 2022-09-16
Validation zhuzh-up: PT speed computation

KasiaKoz:
Resolves: https://github.com/arup-group/genet/issues/145 - Add speed computation and checks

- Adds a method to get a DataFrame with routes : speed (where the speeds are between stop pairs: crowfly distance times a network factor that defaults to 1.3)
- Adds a method to get average speed for each route (average between stop-to-stop pairs)
- Adds checks for 0 and infinity values of speeds in the validation report
- Adds a geojson/shp file output with stop-to-stop speeds in standard outputs, example pic below:
Screenshot 2022-09-15 at 09 17 49

Nb. Part of a wider validation upgrade series. I'm leaving updating the notebook: [6.1. Validating Network: MATSim Specific](https://github.com/arup-group/genet/blob/main/notebooks/6.1.%20Validating%20Network:%20MATSim%20Specific.ipynb) to last in the series.

### DIFF
--- a/genet/schedule_elements.py
+++ b/genet/schedule_elements.py
@@ -260,6 +260,45 @@ class ScheduleElement:
         return None
 
     @abstractmethod
+    def trips_with_stops_to_dataframe(self) -> pd.DataFrame:
+        pass
+
+    def trips_with_stops_and_speed(self, network_factor=1.3) -> pd.DataFrame:
+        """
+        DataFrame: trips_with_stops_to_dataframe, but with speed in metres/second between each of the stops.
+        Note well:
+         - The unit of metres is not guaranteed - this assumes the object is in local metre-based projection.
+         - It does not consider network routes, uses 1.3 network factor as default
+        :param network_factor: Does not consider network routes, network factor (default 1.3) is applied to Euclidean
+            distance.
+        :return:
+        """
+        df = self.trips_with_stops_to_dataframe()
+        df['distance'] = df.apply(
+            lambda row: spatial.distance_between_s2cellids(
+                self._graph.nodes[row['from_stop']]['s2_id'],
+                self._graph.nodes[row['to_stop']]['s2_id']
+            ), axis=1
+        ) * network_factor
+        df['time'] = (df['arrival_time'] - df['departure_time']).dt.total_seconds()
+        df['speed'] = df['distance'] / df['time']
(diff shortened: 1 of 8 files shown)


genet/schedule_elements.py
```diff
@@ -259,6 +259,60 @@ def find_epsg(self):
...
+    def route_attribute_data(self, keys: Union[list, str], index_name: str = None):
+        pass
+
+    @abstractmethod
+    def stop_attribute_data(self, keys: Union[list, str], index_name: str = None):
+        pass
+
+    @abstractmethod
+    def trips_with_stops_to_dataframe(self, gtfs_day='19700101') -> pd.DataFrame:
+        pass
```
KasiaKoz, line 276: All of these methods used to exist just for the schedule, putting them here in the base class `ScheduleElement` (as empty abstract methods) forces all the other classes that depend on it (`Route` and `Service`) to have a version of those methods too. These methods are actually not useful for this PR (oops) but at I thought they were... and having done the work to move them out, I decided to leave it, I think it might some in handy to have them for routes and services 🤷 😄

elizabethc-arup (approved):
This is a really quick and effective way to visualise and validate the speeds of the schedule :D  I had a few questions in regards to the network factor:
- Why is it used instead of the actual length of the route / links? 
- Would the idea be to adjust the network factor based on how windy the roads are in an area?

KasiaKoz:
> Why is it used instead of the actual length of the route / links?

The schedule object doesnt have any information about the network links it's referring to, but mostly laziness. This is a very valid point and I updated the PR with extra functionality. If you pass the network links geodataframe, the method will calculate the routed, network distance and use it for a routed speed. it will return both speeds, because I think each has its' benefits to be looked and limitations.
It's baked into standard outputs to do this automatically and output both speeds as geojsons (/shp files if requested)

Example routes speed geojson (highlights a bad speed for a bus due to a bad snap)
Screenshot 2022-09-26 at 17 26 07

> Would the idea be to adjust the network factor based on how windy the roads are in an area?

Yeah I guess different locations will have a different target for a network route to straight line distance ratio, so I left it open to give the user flexibility :)

elizabethc-arup (approved):
This looks really good! I especially like the idea of having both the options of using the network factor and the route distance to calculate speed which could also be handy for quick visual validations. :D

PR #150, opened 2022-10-07
Fix the release name files generated from the CI build to reflect the…

mfitz:
… default branch name change from master to main


KasiaKoz (approved):
Thank you!!

PR #151, opened 2022-10-12
Use multiprocessing in the notebooks smoketest

mfitz:
Exactly the same as [LAB-1905](https://arupdigital.atlassian.net/browse/LAB-1905).

### Before

Nearly 4.5 minutes running locally
Screen Shot 2022-10-12 at 22 22 47

Around 4 minutes running in the CI build
Screen Shot 2022-10-12 at 22 36 00

### After

Around 1m 15s running locally
Screen Shot 2022-10-12 at 22 27 09

Around 2 minutes running in the CI build
Screen Shot 2022-10-12 at 22 42 52

### Question

The notebooks now run in parallel and finish in a non-deterministic order. They appear in the summary table printed to console at the end of the tests in the (non-deterministic) order that they finished .  _Technically_ it was always the case that they appeared in order of finish, but that was also always name alphabetical order, because we sorted the list and then worked through it sequentially in series. 

Do we care that the table is no longer alphabetically sorted? It's a quick fix if we do.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,4 @@ scripts/secrets-mgr-exploration.py
 example_data/pt2matsim_network/genet_output/*
 genet_output/*
 example_data/output_*/*
+example_data/api_requests_send.json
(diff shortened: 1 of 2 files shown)


KasiaKoz (approved):
nice one, I don't care about the order of the notebooks in the output table :)

PR #153, opened 2022-10-24
Split (active) modal subgraphs

KasiaKoz:
Adds a script to split given modes from the links on the network graph from under the `'modes'` attribute. You end up with a graph with many more links, the links with mentioned modes exists only with that mode, all other attributes are copied across (for the time being anyway, we might have to set/scale capacity/speed eventually).

Context: We use Hermes in MATSim, which results in slower vehicles, like bikes, blocking cars on links. Creating separate links for that mode is an initial solution to that problem.

Bonus: Noticed a lag with method that adds links. Because with this script we're adding links with the same from and to nodes, we need to generate a multi-index (links are stores in the graph in this format: `(from_node, to_node, multiindex)`). One line, which gives a closer initial guess for the multi-index speeds things up for larger graphs:

Screenshot 2022-10-24 at 10 34 20
(time in seconds)

### DIFF
--- a/scripts/separate_modes_in_network.py
+++ b/scripts/separate_modes_in_network.py
@@ -1,0 +1,75 @@
+import argparse
+import json
+import logging
+import os
+
+from genet import read_matsim
+from genet.utils.persistence import ensure_dir
+from genet.output.sanitiser import sanitise_dictionary
+
+if __name__ == '__main__':
+    arg_parser = argparse.ArgumentParser(description='Generate own links for given modes in a MATSim network, '
+                                                     'creating a separate modal subgraph.')
+
+    arg_parser.add_argument('-n',
+                            '--network',
+                            help='Location of the network.xml file',
+                            required=True)
+
+    arg_parser.add_argument('-p',
+                            '--projection',
+                            help='The projection network is in, eg. "epsg:27700"',
+                            required=True)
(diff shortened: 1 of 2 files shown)


scripts/separate_modes_in_network.py
```diff
@@ -0,0 +1,77 @@
...
+        update_mode_links = {k: {'modes': df.loc[k] - {mode}} for k in modal_links}
+        new_links = {f'{mode}---{k}': {**n.link(k), **{'modes': {mode}, 'id': f'{mode}---{k}'}} for k in modal_links}
+        n.apply_attributes_to_links(update_mode_links)
+        n.add_links(new_links)
+
+    logging.info(f'Number of links after separating graph: {len(n.link_id_mapping)}')
+
+    n.write_to_matsim(output_dir)
+
+    logging.info('Generating validation report')
```
mfitz, line 65: I like the automatic validation of the network following the change 👍

mfitz (changes requested):
Looks good. Might benefit from a little more of an explanation about what the script is for and what it does in the help screen.

KasiaKoz:
Improved script description

fredshone (approved):

mfitz (approved):

Issue #122, opened 2022-06-17
Automatically generate length for added links


KasiaKoz:
Once we make spatial information for nodes non-optional: https://github.com/arup-group/genet/issues/121 we should default to genet calculating straight line (or not straight, if geometry is being added too) distance for links being added, for convenience.

PR #161, opened 2022-11-23
Automatically calculate length when adding new link(s)

ana-kop:
PR for issue: https://github.com/arup-group/genet/issues/122


ana-kop:
So far have only changed the method `add_link()` to add an optional length calculation, but probably need to also:
1) Add this option to the `add_links()` method
2) Make it default rather than an option? (would require a large number of unit tests to be corrected...)

PR #170, opened 2023-04-04
Elevation script fixes

KasiaKoz:
Fixes:
- clashing argument flags `-ws`
- `import genet.utils.elevation as elevation` being overwritten by `elevation = args['elevation']` and preventing [`write_slope_to_object_attribute_file`](https://github.com/arup-group/genet/compare/elevation-script-fixes?expand=1#diff-99b2946aac5db3dd9ee2030e299fe866323e2dc3d2ed02fd93e16d1d066708edL142-R144)

As a bonus:
- gets rid of `elevation_output_dir = os.path.join(output_dir, 'elevation')` it's a needless extra folder (nothing gets saved anywhere else). User can decide if they want outputs saved to a folder called elevation or not

### DIFF
--- a/genet/__init__.py
+++ b/genet/__init__.py
@@ -8,3 +8,4 @@ from genet.exceptions import ScheduleElementGraphSchemaError, RouteInitialisatio
     ServiceInitialisationError  # noqa: F401
 from genet.utils import graph_operations  # noqa: F401
 from genet.utils import google_directions  # noqa: F401
+from genet.utils import elevation  # noqa: F401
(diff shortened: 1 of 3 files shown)


mfitz (approved):
Looks good to me.

I think we say this every time, but it feels like we could do with some testing around the functionality in these scripts (which I think we've also talked about collecting into a CLI, right?).

KasiaKoz:
>I think we say this every time, but it feels like we could do with some testing around the functionality in these scripts (which I think we've also talked about collecting into a CLI, right?).

yeah, it's been on the (jira) cards for a while (unfortunately I'm struggling to find the card right now - I already started matching up some example inputs to the scripts on the card)

PR #173, opened 2023-04-25
Fix merging lists in complex dictionaries and `add_services` from other `Schedule`s

KasiaKoz:
The method used for merging complex dictionaries (dictionaries that have list, sets or other dictionaries of that complex type as values) has been messing up the lists.

If I have two dictionaries with lists as values, like so:
```python
a = {'key': [1,2,3]}
b = {'key': [1,2,3]}
```
they would have been merged to give:
```python
{'key': [1,2,3]}
```
or any other order of that list (`[2, 1, 3]`, `[3, 2, 1]` etc.), because the method was converting the lists to sets, combining them with `OR`, and then calling a list on the output set. The order of the list would not be guaranteed to be preserved. This came up as a problem when merging services with ordered stops - the order is super important and some stops could repeat themselves (nothing wrong with repetitions - converting to set makes all items unique in the set, altering the route even more).

So this PR fixes that. The output of merging `a` and `b` is now:
```python
{'key': [1,2,3,1,2,3]}
```
the lists are merged together and order preserved. Now, this threw off some methods that merge dictionaries that index schedules - which were already broken! (the order was lost, now the stops and a bunch of other lists would grow and grow).

Here is a test catching the bug on both the main branch:
Screenshot 2023-04-25 at 15 16 48

and this branch before fixing it, but after the change to the dictionary merging method:
Screenshot 2023-04-25 at 15 16 17

after the fix:
Screenshot 2023-04-25 at 16 22 45

Potentially bad, but I had a think about which of our networks may have suffered negative outcomes from this bug and I don't think any of them are at risk. It appeared in a really particular setup/usage which was not used by any of the PT network scenarios I am aware of. The particular usage/setup was:
- adding multiple services via the `add_services` method of the network schedule, AND
- the services being added were part of another schedule object (this happens when we add PT using a GTFS feed for example)

(so the only network that would be at risk is Ireland, as far as I know, but a different method (`add`) is used to add two schedules rather than `add_services`; plus, going through the data in one of the scenario networks, it could not have been through a `set->list` operation since duplicates exist in the fields that would have been affected)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM python:3.7-slim-stretch
+FROM python:3.7.16-bullseye
 
 RUN apt-get update && \
 apt-get upgrade -y && \
-apt-get -y install gcc git libspatialindex-dev curl coinor-cbc &&  \
+apt-get -y install gcc git libgdal-dev libgeos-dev libspatialindex-dev curl coinor-cbc &&  \
 rm -rf /var/lib/apt/lists/*
 
 RUN curl -sL https://deb.nodesource.com/setup_17.x | bash -


tests/test_core_schedule.py
```diff
@@ -872,6 +872,22 @@ def test_multiple_services_are_present_in_schedule_after_adding(schedule, servic
...
 
 
+def test_adding_services_from_schedule_does_not_change_route_data(
+        schedule, services_to_add):
+    routes_data_before_add = [r.__dict__ for r in services_to_add[0].routes()]
+    schedule_to_add = Schedule(services=services_to_add, epsg='epsg:27700')
+
+    schedule.add_services(list(schedule_to_add.services()))
+
+    for item in ['ordered_stops', 'arrival_offsets', 'departure_offsets']:
```
mfitz, line 882: Can we simplify this somehow? Ideally, I'd want to be able to do something like:

```python
routes_before_add = schedule.routes

schedule.add_services(services_to_add) # or schedule.add_services(schedule_to_add)

assert schedule.routes == routes_before_add 

```

If we don't want to change the API, maybe we can use test helper methods so we still get something like:

```python
assert_route_data_equal(schedule.routes, routes_before_add)
```

mfitz (changes requested):
I've left inline comments, some of which are questions, others suggestions. We might want a couple more tests down at the level of the merging function, and maybe a clearer test of the higher level functionality around adding services.

tests/test_utils_dict_support.py
```diff
@@ -89,6 +89,24 @@ def test_merging_dictionaries_with_lists():
...
 
+def test_merging_dictionaries_with_lists_with_non_unique_items():
+    return_d = dict_support.merge_complex_dictionaries(
+        {'b': [6, 6]},
+        {'b': [5]}
+    )
+    assert_semantically_equal(return_d, {'b': [6, 6, 5]})
+
+
+def test_merging_dictionaries_with_lists_with_non_unique_items_does_not_change_input_dicts():
```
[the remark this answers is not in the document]
KasiaKoz: it's not unique to lists, i've added more to this test and changed the name to make it general

tests/test_utils_dict_support.py
```diff
@@ -89,6 +89,24 @@ def test_merging_dictionaries_with_lists():
     assert_semantically_equal(return_d, {'a': 1, 'b': [3, 6, 5], 'c': [1, 8, 90]})
 
 
+def test_merging_dictionaries_with_lists_with_non_unique_items():
+    return_d = dict_support.merge_complex_dictionaries(
+        {'b': [6, 6]},
+        {'b': [5]}
+    )
+    assert_semantically_equal(return_d, {'b': [6, 6, 5]})
```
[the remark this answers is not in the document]
KasiaKoz: added a couple more tests for these cases

tests/test_utils_dict_support.py
```diff
@@ -119,12 +137,6 @@ def test_merging_dicts_with_lists():
     assert_semantically_equal(d, {'1': ['', '2'], '2': [], '3': ['1']})
 
 
-def test_merging_dicts_with_lists_with_overlapping_values_returns_list_with_unique_values():
```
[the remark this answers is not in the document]
KasiaKoz: Yeah, i had enough sense to know how it works but not how it could yield bad results. And didn't feel weird to me at the time to covert a list to set and back to list for some reason :shrug:

tests/test_core_schedule.py
KasiaKoz: I can simplify it to:
```
    schedule_to_add = Schedule(services=services_to_add, epsg='epsg:27700')
    route_ids_to_add = list(schedule_to_add.route_ids())
    routes_data_before_add = [schedule_to_add.route(r) for r in route_ids_to_add]

    schedule.add_services(list(schedule_to_add.services()))

    assert [schedule.route(r) for r in route_ids_to_add] == routes_data_before_add
```
but that does not compare the time offsets and trips. The things that are compared are only the list items inside the routes, so it would not be a generic `assert_route_data_equal` that could be reused, it would be `assert_route_list_data_equal` and used only once in this test.

mfitz: I know what the test is doing in general terms, but I find the detail of the assertions confusing. There is implicit knowledge of the fixtures being used in the assertions, e.g. `'new_route_0_0'` - nothing in the test reveals why we expect what we expect in the assertions.

The test should be more straightforward: grab the routes before, call the code under test, and compare the routes now to the routes before. That shouldn't require us to know anything about the actual _content_ of the routes. We don't care about the specifics; we only care that they haven't changed. We're doing work in the test that doesn't belong there.

The `Route` class has an equality comparator that you've alluded to. If it's missing some of the things that make a comparison between routes useful (time offsets and whatnot), should we add those? That feels like the place where the mechanics of route equality comparison should live.

tests/test_utils_dict_support.py
```diff
@@ -89,6 +89,64 @@ def test_merging_dictionaries_with_lists():
...
+        'c': {
+            'a': {2, 3},
+            'b': [5]
+        },
+        'd': 'yo'
+    }
+
+    return_d = dict_support.merge_complex_dictionaries(A, B)
+
+    assert_semantically_equal(A, {'a': {1, 2}, 'b': [6, 6], 'c': {'a': {1, 2}, 'b': [6, 6]}, 'd': 'hey'})
```
mfitz, line 146: Could you use deep copies of the two dicts taken before the call to merge as your "befores" and then compare `A` and `B` to their corresponding "before" states?

mfitz (changes requested):
All good, but I have more questions about `test_adding_services_from_schedule_does_not_change_route_data`

tests/test_core_components_route.py
```diff
@@ -205,9 +205,11 @@ def test_routes_equal(stop_epsg_27700):
     b = Route(
         route_short_name='route', mode='bus',
         stops=[stop_epsg_27700, stop_epsg_27700],
-        trips={},
-        arrival_offsets=[],
-        departure_offsets=[])
+        trips={'trip_id': ['VJ00938baa194cee94700312812d208fe79f3297ee_04:40:00'],
```
mfitz, line 208: 👍

tests/test_core_schedule.py
```diff
@@ -872,6 +871,17 @@ def test_multiple_services_are_present_in_schedule_after_adding(schedule, servic
     assert set(schedule.route_ids()) == expected_routes_after_adding
 
 
+def test_adding_services_from_schedule_does_not_change_route_data(
```
mfitz, line 874: Nice!

mfitz (approved):

PR #179, opened 2023-07-12
MATSim sprinkle scripts

KasiaKoz:
Adds new and modifies existing scripts to bring you a suite of scripts with our most common alterations to MATSim networks. These are used in our Grand Net Auto workflows. 
The future location/setup of all genet scripts is uncertain - @brynpickering made good points on this, and they might move somewhere else one day.

# Add Elevation [changes]

Small change to output visuals into a `supporting_outputs` subfolder to match other scripts.

# Auto Schedule Fixes [new, and additions to genet codebase]

This is the biggest change. It takes Anastasia's work on the Sheffield project - [internal CML project documentation page](https://github.com/arup-group/sheffield-network/blob/main/7_schedule_fixes/Fix_zero_headways_and_infinite_speeds.ipynb) - and writes it into a few new genet methods for checking and fixing zero headway and infinite speeds in the PT schedules. A script calls on those methods and generates some supporting visual outputs for before and after.

### Zero Headways
In our networks, each PT service has unique trips. Headway is calculated as the difference between trip departure times. We can look at different statistics of headway minutes of services. In this step we check that there are no 0 minute headways - that would mean two of the same trips start at the same time. To correct this we delete one of the trips, treating it as a duplicate.

The script outputs a geojson of the PT network with headway stats before and after the fix. Below we look at headway minimum values between values 0 and 1 minute to see the 0 headways disappear.

Auto PT Schedule Fix - Zero Headways

### Infinite Speeds
In our networks we know the times each PT vehicle arrives and leaves a stop. We calculate speed between each stop pair for services. We use the declared times at stops (which are inherited from GTFS) and crow-fly distance * 1.3 network factor (this is done also for routed modes like bus to simplify things, as we're only after infinite speeds which will show up whether we use the true route or not). Infinite speeds exist as a consequence of division by zero (time). This is corrected by recalculating the arrival and departure times at the problem stops - the times at other stops are kept the same as much as possible.

The script outputs a geojson of the PT network with maximum speeds along unique routes of PT services. Note, the infinite speeds are replaced by large 9999 value to help with visualisation of these values.

Auto PT Schedule Fix - Infinite Speeds

# Separate Modes [changes]

This script exists to support current active mode implementation in MATSim with hermes. The changes here are to generate supporting visual outputs.

extracting walking graph

# Simplification [changes]

Added a param to force connecting disconnected subgraphs, this approach is used on another CML project. There can be legitimate reasons for the network graph to be disconnected (multiple islands), but MATSim expects a fully, strongly connected graph for each routed mode. If you are expecting a disconnected graph, you can set this parameter to `True` and connecting links with capacity and speed penalties will be added to the network.

# Squeeze External [new script]

Script to reduce capacity and speed outside of the study area (as done on many projects now, e.g. [internal CML project documentation page](https://github.com/arup-group/sheffield-network/blob/main/8_network_squeeze/2_Reduce_capacity_speed_outside_study_area.ipynb))

External Squeeze only

# Squeeze Urban [new script]

Script to reduce capacity and speed outside of the study area (as done on many projects now, e.g. [internal CML project documentation page](https://github.com/arup-group/sheffield-network/blob/main/8_network_squeeze/1_Rural_urban_tagging_and_urban_squeeze.ipynb).
It has an option to include the study area subsetting - which means it can be used in tandem with external script - or not, just to use on it's own.

Urban minor road squeeze
Urban Squeeze within study area

### Bonus pic if you're wondering what both squeezes look like together
Network Squeeze before and after

# Validate Network [changes]

Small fix to sanitise the validation report dictionary before saving to json file.

# Scale Vehicles [new script]

Uses @elizabethc-arup vehicle scaling functionality in a standalone script. This makes it an independent step that can be run on any network in a workflow.


elizabethc-arup (approved):
Exciting changes, particularly the schedule fixes would be really helpful. Have come across a few speed and 0 headway warnings form the genet validation of a project network which wasn't very straight forward to pin down and fix.

PR #185, opened 2023-08-22
Release v4.0.0

KasiaKoz:
It's been over a year since last release, oops. Loads have changed so I think it's time for a new release! ✨

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,36 @@
 # Changelog
 
-## [Unreleased]
+## [v3.1.0] - 2023-08-
 
 ### Added
 * Validation for [intermodal access/egress for MATSim](https://github.com/matsim-org/matsim-libs/tree/master/contribs/sbb-extensions#intermodal-access-and-egress)
+* GeNet Scripts for: [#179](https://github.com/arup-group/genet/pull/179)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,36 @@
 # Changelog
 
-## [Unreleased]
+## [v3.1.0] - 2023-08-
```
mfitz, line 3: The version number in `setup.py` doesn't seem to have been updated?

mfitz (changes requested):
Looks mostly there, but the version number in `setup.py` needs to be in sync.

What's the thinking behind this being a minor version release? I'm not suggesting it's a major release, but I want you to show your working out 😄

KasiaKoz:
>What's the thinking behind this being a minor version release? I'm not suggesting it's a major release, but I want you to show your working out 😄

I changed it to a major one now, I realised one of the changes is a breaking one - nodes need spatial information now, so old code without them will break. I thought, even though it's a lot of stuff, none of them were breaking changes, but there is one, no matter how stupid it actually is (there is no point adding nodes without spatial information as that would break matsim / make no sense)..

mfitz (approved):

PR #188, opened 2023-09-21
Restructure project; update to py311.

brynpickering:
Reprojection tests are all failing as everything has moved O(~1m) on going from lat/lon to x/y - M1 mac issue? New pyproj/geoapandas/gdal/fiona version issue? Hoping CI will shed some light.

PR #190, opened 2023-09-28
Add scripts integration tests

KasiaKoz:
Up until now the example scripts in genet were not tested. This adds integration style tests, which run the scripts with example input data and checks the most vital expected files are generated. These don't check the contents of the files - any code within those scripts should be mostly trivial, leaning mostly on genet methods which are unit tested. All but one of the scripts are expected to complete successfully. The google directions API script will fail with exceeded threshold of api requests.

This is a long time coming, finally completing [LAB-1185](https://arupdigital.atlassian.net/browse/LAB-1185) and required by changes which move scripts to cli here: https://github.com/arup-group/genet/pull/188 to verify if they have been ported correctly.

There are some bonus corrections of doc and help strings here too.

[LAB-1185]: https://arupdigital.atlassian.net/browse/LAB-1185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/auto_schedule_fixes.py
```diff
@@ -33,7 +33,18 @@ def generate_speed_geojson(n, gdf, output_dir, filename_suffix):
 
 if __name__ == '__main__':
     arg_parser = argparse.ArgumentParser(
-        description=''
+        description='Script to check and correct, if needed, the speed and headway of services in the schedule.'
```
brynpickering, line 36: I prefer descriptions to break over fewer lines (see changes in #188). This will be automaticall wrapped when calling `--help` in the CLI anyway. I would break only where natural. You can also use `""" ... """` to do it all as one big string block. It's then easier to read when maintaining.

tests/test_scripts.py
```diff
@@ -0,0 +1,310 @@
...
+
+    exit_code, console_output = run_process(command)
+
+    if exit_code != 0:
+        raise AssertionError(f'Script {script} did not run successfully. Error message: \n\n{console_output}')
+    else:
+        check_output_files_exist(output_folder, SCRIPTS[script]['expected_files'])
+
+
+def test_google_api_script_throws_error_without_api_key(tmpdir):
```
brynpickering, line 285: could we have a test that runs the script with an API key that's stored in the repository secrets? If we make one request per CI run, would that keep us easily below some threshold for being charged for the requests?

tests/test_scripts.py
```diff
@@ -0,0 +1,310 @@
...
+        ]
+    }
+    output_folder = os.path.join(tmpdir, f'script_{script.replace(".py", "")}')
+    check_output_files_do_not_exist(output_folder, script_info['expected_files'])
+    command = f"python {os.path.join(os.path.dirname(__file__), '..', 'scripts', script)} "
+    command = command + ' '.join(script_info['args']) + ' --output_dir ' + output_folder
+
+    exit_code, console_output = run_process(command)
+
+    assert exit_code != 0, f'Script {script} is expected ot fail.'
```
brynpickering, line 307: ot -> to?

brynpickering (commented):
Some minor comments. A bigger one: do you want this merged into current main? Once integrated into #188, the tests will look quite different (using Click runners). Just wondering if it's worth doing that now and making it a PR request onto the #188 branch or merging this into current main and then making the changes necessary to fit with #188. Up to you!

example_data/Fitzrovia_urban_polygon.geojson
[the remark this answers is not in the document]
KasiaKoz: there are some test tif files, this one overlays with the example network in that folder

genet/utils/google_directions.py
```diff
@@ -18,7 +18,7 @@
 
 def send_requests_for_network(n, request_number_threshold: int, output_dir,
                               departure_time, traffic_model: str = None,
-                              max_workers: int = 4, key: str = None, secret_name: str = None, region_name: str = None):
+                              key: str = None, secret_name: str = None, region_name: str = None):
```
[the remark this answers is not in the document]
KasiaKoz: I will save this typing tip for a larger overhaul of genet methods (logged an issue to capture it: https://github.com/arup-group/genet/issues/191), this one method is just a drop in an ocean, I'm just cleaning a dead param that wasn't being used for now.

scripts/auto_schedule_fixes.py
KasiaKoz: ok, I will keep that in mind when I port this to cli in your branch 👍

tests/test_scripts.py
KasiaKoz: so, we could, but I don't want to actually ping the api every time we run tests (it would also require people to have those secrets set up locally if they want to run tests or be authenticated to the aws account). There isn't a free amount of queries btw, I don't think it would break the bank tbh. There are unit tests that mock querying the api and processing the results, I really think that testing this script breaking on t'hreshold limit of requests exceeded' before querying the api is more valuable for us here.

That said, I think in the cli version in your branch, we might be able to mock this and run without pinging the api.

KasiaKoz (commented):
> do you want this merged into current main?
Yeah, I want to merge this to main and have a pretty straightforward history for this one. I will get to work on porting this to your branch as soon as we get this merged 👍

brynpickering (approved):

PR #194, opened 2023-10-12
Clickify scripts

brynpickering:
Moved CLI into genet source code. I realise this has scope to break lots of things in GNA, but seems to be a better long-term solution to what are the main interface methods to the source code.

Some geospatial checks fail again locally, but I'm hoping the CI will say it's all clean.

I've added some comment/docstring line length NOQAs. These will be removed once moving to linting with Ruff (where you can set a different allowed length for code and comments/docs). 

Slowly moving towards using `pathlib.Path` over `os` btw. The Path objects are generally safer (and cleaner) to work with and you can get click to automatically parse an input as a Path, which is neat.


genet/cli.py
```diff
@@ -0,0 +1,1504 @@
...
+@cli.command()
+@xml_file("network")
+@xml_file("schedule", False)
+@xml_file("vehicles", False)
+@projection
+@vehicle_scalings
+@output_dir
+def auto_schedule_fixes(
+    path_to_network: Path,
+    path_to_schedule: Optional[Path],
```
KasiaKoz, line 443: this should be optional really, I don't know what I was thinking in the original script - without a schedule there is nothing to fix.. could you update this to be non optional?

genet/variables.py
```diff
@@ -51,3 +51,5 @@
 # e.g. {1: "subway", 400: "subway", 401: "subway", 402: "subway", ...]}
 EXTENDED_TYPE_MAP = dict(zip(new_keys, new_values))
 del new_keys, new_values
+
+EPSG4326 = "epsg:4326"
```
KasiaKoz, line 55: nice

tests/test_cli.py
```diff
@@ -0,0 +1,343 @@
...
+PROJECTION = 'epsg:27700'
+
+@pytest.fixture(scope="function")
+def invoke_runner_and_check_files(tmpdir_factory):
+    outdir = tmpdir_factory.mktemp("outdir")
+
+    def _invoke_runner_and_check_files(
+        func: str, args: list[str], expected_files: list = [], not_expected_files: list = []
+    ):
+        runner = CliRunner()
```
KasiaKoz, line 24: in the original scripts tests there was a check before running the process that the expected files do not already exist, can you add it here?

tests/test_cli.py
```diff
@@ -0,0 +1,343 @@
...
+            args= [
+                f'--network={EXAMPLE_NETWORK}',
+                f'--projection={PROJECTION}',
+                '--requests_threshold=1',
+                '--subset_conditions=primary,motorway',
+                '--traffic_model=best_guess',
+                '--departure_time=now'
+            ],
+            not_expected_files=[
+                'api_requests.json',
```
KasiaKoz, line 338: the expectation is that the requests will be generated and saved even if sending them fails

KasiaKoz (changes requested):
Good attempt, a few changes needed imo. A big one that isn't covered by my comments is removing the scripts directory to remove duplication and updating the readme

KasiaKoz (approved):

PR #197, opened 2023-10-24
Update `Dockerfile` and docs

KasiaKoz:
PR to https://github.com/arup-group/genet/pull/194 which converts all scripts to CLI.

This PR adds changes to `Dockerfile` to support this - I went with no entry point. My reasoning is that it makes it flexible to run both genet CLI and custom scripts without overriding the entry point - which is a bit of a headache if using an AWS batch job for example.

If the docker image is named something else, like 'cml-genet' then it doesn't look too crazy I think... (see the docs) - but I'm  looking for feedback on this decision.

Docker image is building: [slack build notification](https://arupconsulting.slack.com/archives/CNASXCHM2/p1698156704065569), and tests inside passing
```
$ docker run cml-genet python -m pytest -vv tests
...
=========== 1063 passed, 1 xfailed, 82 warnings in 279.00s (0:04:38) ===========
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,33 +61,41 @@ Docker is the recommended way to use GeNet if you do not plan to make any code c
 
     docker build -t "genet" .
 
-#### Running a container with a pre-baked script
+#### Using the cli inside a container
 
-    docker run genet reproject_network.py -h
-    
-    usage: reproject_network.py [-h] -n NETWORK [-s SCHEDULE] [-v VEHICLES] -cp
-                                CURRENT_PROJECTION -np NEW_PROJECTION
-                                [-p PROCESSES] -od OUTPUT_DIR
-    
-    Reproject a MATSim network
-    
-    optional arguments:
-      -h, --help            show this help message and exit
-      -n NETWORK, --network NETWORK
-                            Location of the network.xml file
-      -s SCHEDULE, --schedule SCHEDULE
-                            Location of the schedule.xml file
-      -v VEHICLES, --vehicles VEHCILES
-      							  Location of the vehicles.xml file
-      -cp CURRENT_PROJECTION, --current_projection CURRENT_PROJECTION
-                            The projection network is currently in, eg.
-                            "epsg:27700"
-      -np NEW_PROJECTION, --new_projection NEW_PROJECTION
-                            The projection desired, eg. "epsg:27700"
-      -p PROCESSES, --processes PROCESSES
-                            The number of processes to split computation across
-      -od OUTPUT_DIR, --output_dir OUTPUT_DIR
-                            Output directory for the reprojected network
+    docker run genet --help
+
+to show the list of available commands, and e.g.
+
+    docker run genet simplify-network --help
+
+to show description of the command and parameters:
+```commandline
+Usage: genet simplify-network [OPTIONS]
+
+  Simplify a MATSim network by removing intermediate links from paths
+
+Options:
+  -n, --network PATH              Location of the input network.xml file
+                                  [required]
+  -s, --schedule PATH             Location of the input schedule.xml file
+  -v, --vehicles PATH             Location of the input vehicles.xml file
+  -p, --projection TEXT           The projection network is in, eg.
+                                  "epsg:27700"  [required]
+  -pp, --processes INTEGER        Number of parallel processes to split
+                                  process across
+  -vsc, --vehicle_scalings TEXT   Comma delimited list of scales for vehicles
+  -od, --output_dir DIRECTORY     Output directory  [required]
+  -fc, --force_strongly_connected_graph
+                                  If True, checks for disconnected subgraphs
+                                  for modes `walk`, `bike` and `car`. If there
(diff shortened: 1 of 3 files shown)


mfitz (commented):
How big is the Docker image? Is it worth adding it to the GitHub release as an asset?

KasiaKoz:
> How big is the Docker image? Is it worth adding it to the GitHub release as an asset?

@mfitz it's very big: `6.14GB` 
I have a genet image from 10 months ago that was `1.52GB`, I'm not sure what inflated the size so much (this PR: https://github.com/arup-group/genet/pull/174 brought it up to ~2gig, not sure where the rest came from, maybe we're copying jupyter notebooks and those were rerun had all the widgets saved... ? I'm only speculating here)

It would be cool to add it as an asset, but only if we trimmed the weight a bit I think

mfitz:
> > How big is the Docker image? Is it worth adding it to the GitHub release as an asset?
> 
> @mfitz it's very big: `6.14GB` I have a genet image from 10 months ago that was `1.52GB`, I'm not sure what inflated the size so much (this PR: #174 brought it up to ~2gig, not sure where the rest came from, maybe we're copying jupyter notebooks and those were rerun had all the widgets saved... ? I'm only speculating here)
> 
> It would be cool to add it as an asset, but only if we trimmed the weight a bit I think

Hmm. When I build it locally from this branch it's 2.8 GB. Is that 6 GB IMAGE your local build, or the one in ECR?

mfitz (changes requested):
The version number doesn't seem to have changed - can we do that?

brynpickering:
It won't be the notebooks, they'd never get big enough. I would expect the image to be smaller once we reach `mamba` nirvana (it proved to be the case for PAM). Is there precedence of projects releasing docker images as GH assets @mfitz ?

README.md
```diff
@@ -160,9 +193,14 @@ described below to verify that the code still works. All of the following comman
 directory.
 
 ### Unit tests
+To run unit tests within genet python environment:
 
     python -m pytest -vv tests
 
+and within a docker container:
+
+    docker run cml-genet python -m pytest -vv tests
```
brynpickering, line 202: does `docker run cml-genet pytest -vv` have the same effect?

mfitz:
> Is there precedence of projects releasing docker images as GH assets @mfitz ?

Not in one of our own projects @brynpickering. I think we can consider publishing the docker image, though, and I can think of 3 ways to do that:

- make the GeNet ECR repo public
- publish the image to GitHub Packages
- attach the image to GitHub releases as an asset

I don't think it's a burning need, but I think it would help drive good behaviour around versioning and releases.

KasiaKoz:
>Hmm. When I build it locally from this branch it's 2.8 GB. Is that 6 GB image your local build, or the one in ECR?

LOL! I have an ignored folder in there that has all the outputs from all the scripts after running them on londinium... I think that'll do it 😆

brynpickering:
btw, would a publicly available genet docker image become redundant if genet was available for install directly from pip/conda?

README.md
KasiaKoz: `docker run cml-genet pytest -vv tests` does, will change

KasiaKoz:
the docker image serves a different purpose in my mind, you need it for containerised workflows like we use on aws---I don't know if anyone else is using it with docker (or using genet at all tbh) but personally I'm always really happy when I see a published docker image for something I want to try adding to a pipeline. Having said that, I think sharing a working `Dockerfile` is already really good.

mfitz:
> the docker image serves a different purpose in my mind, you need it for containerised workflows like we use on aws---I don't know if anyone else is using it with docker (or using genet at all tbh) but personally I'm always really happy when I see a published docker image for something I want to try adding to a pipeline. Having said that, I think sharing a working `Dockerfile` is already really good.

Plus, the README tells people to use the Docker image, so if we're pushing that as the preferred way to use GeNet, it might be nice to have a pre-built, clearly versioned image somewhere. It's not difficult to clone the repo and build it yourself, of course, but my agenda here is making release/versioning a first-class citizen 😄 

@brynpickering If we replaced Docker as the recommended way to use GeNet with pip/conda install, the release/versioning question would take care of itself, I think, so that would solve my particular problem, but as @KasiaKoz says, we'd still need the Docker image for AWS Batch.

KasiaKoz:
@mfitz I added notes to the changelog rather than changing the version (i didn't quote this PR though as it's a small part of another) - is that acceptable in your opinion? i don't think we want to do a release yet, we'll definitely do one after we reach the mamba nirvana

mfitz:
> @mfitz I added notes to the changelog rather than changing the version (i didn't quote this PR though as it's a small part of another) - is that acceptable in your opinion? i don't think we want to do a release yet, we'll definitely do one after we reach the mamba nirvana

Cool @KasiaKoz 👍

mfitz (approved):

brynpickering (approved):

PR #203, opened 2023-10-31
Refine cli

KasiaKoz:
Moved genet and non-cli code out of the main cli module.

On the side - the genet help command feels a bit slow. I thought maybe moving and tidying up the imports would speed it up, it did a little but not as much as I hoped.


genet/use/cli.py
```diff
@@ -0,0 +1,951 @@
...
+                f'Schedule vehicle level validation: {report["schedule"]["vehicle_level"]["vehicle_definitions_valid"]}'
+            )
+        logging.info(
+            f'Routing validation: {report["routing"]["services_have_routes_in_the_graph"]}'
+        )
+
+    _to_json(report, output_dir / "validation_report.json")
+
+    logging.info("Generating summary report")
+    summary_report = network.summary_report()
```
KasiaKoz, line 50: This is the biggest change to code, summary report is very useful and goes hand in hand with validation of the network

tests/test_cli.py
```diff
@@ -317,7 +317,10 @@ def test_validate_network(invoke_runner_and_check_files):
             f'--vehicles={EXAMPLE_VEHICLES}',
             f'--projection={PROJECTION}',
         ],
-        expected_files=['validation_report.json']
+        expected_files=[
+            'validation_report.json',
+            'summary_report.json',
```
KasiaKoz, line 322: now also expect summary report output when running validation command

genet/use/cli.py
```diff
@@ -0,0 +1,951 @@
+import json
```
mfitz, line 1: A question about naming...

It doesn't seem like the code in this module is tied to the CLI. If I were using GeNet from a notebook, an app with a web front end, or a Python script, (i.e. *not* the CLI) I would be using many of these functions to do things like read or write a network, generate Google Directions API requests, etc.

In [MVC](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) terms, this module is arguably a mixture of **controller** and **model**, but it isn't the **view**/bit used directly by the command line user.

This refactoring is about the separation of concerns, which I like. We've made the CLI code a thin layer around the "business logic" of the library, responsible only for figuring out what the command line user wants to do and translating that into operations on the core library. Good stuff.

But the code in this module has now (rightly) been relieved of all CLI responsibility and represents "business logic", so the name `cli` doesn't seem right to me. There isn't actually any CLI code in here, as far as I can see, and there's no reason this module should be/considered to be for the exclusive use of the CLI.

Can we come up with a better, more descriptive name?

genet/cli.py
```diff
@@ -1,153 +1,12 @@
...
-from genet.output.geojson import (
-    generate_headway_geojson,
-    generate_speed_geojson,
-    modal_subset,
-    save_geodataframe,
-)
-from genet.utils.persistence import ensure_dir
-from genet.variables import EPSG4326
 
 logging.basicConfig(level=logging.INFO, format="%(levelname)-3s %(message)s")
```
mfitz, line 6: Should the logging config live "deeper" inside the library? Would we want it to be consistently configured no matter what route the user is taking (CLI, web app, Python script, notebook, etc.) into the code?

mfitz (changes requested):
I like the separation of concerns and the CLI code now being a separate, thin layer in front of the main logic of the module 👍 

I think the `use.cli` module is misnamed - I've made a longer inline comment explaining my thoughts.

I was also surprised to find almost no changes to test code - why is that? Is most of the code that moved untested?

genet/use/cli.py
KasiaKoz: I'm leaning towards `recipes` - does that sound good?

KasiaKoz:
>I was also surprised to find almost no changes to test code - why is that? Is most of the code that moved untested?

@mfitz It's tested via cli commands, I could add testing the functions as well, but it would be almost exactly the same kind of tests, duplicating things. These are more on the integration side of testing too so not super fast to run.

KasiaKoz:
@mfitz I tidied up the logging config which was in a couple of places, renamed the module and improved testing to separate the testing of recipes code and cli just testing that it calls on the right function with the right arguments.

genet/__init__.py
```diff
@@ -14,6 +16,10 @@
     read_osm,
 )
 from genet.max_stable_set import MaxStableSet
+from genet.output import geojson
 from genet.schedule_elements import Route, Schedule, Service, Stop
 from genet.use.road_pricing import Toll
-from genet.utils import elevation, google_directions, graph_operations
+from genet.utils import elevation, google_directions, graph_operations, spatial
+
+logging.basicConfig(format="%(levelname)-3s %(asctime)s - %(message)s", level=logging.INFO)
```
brynpickering, line 24: For logging, what I would suggest is something like this (based on what we do in [calliope](https://github.com/calliope-project/calliope/blob/3779e0dff95c26141293a528f3ec4416cc292783/src/calliope/core/util/logging.py#L19)):

```python
_time_format = "%Y-%m-%d %H:%M:%S"
root_logger = logging.getLogger("genet")  # Get the root logger

# Remove any existing output handlers from root logger
if root_logger.hasHandlers():
    root_logger.handlers.clear()

# Create a console log handler with decent formatting and attach it
formatter = logging.Formatter(
        "%(levelname)-3s %(asctime)s - %(message)s", datefmt=_time_format
)
console = logging.StreamHandler(stream=sys.stdout)
console.setFormatter(formatter)
root_logger.addHandler(console)
```
Then in any file that you want to use a logger, just use `logger = logging.getLogger(__name__)` and because it is a child of the root logger (e.g., `genet.core` is a child of `genet`, as is `genet.cli`) it will inherit the root logger's handler. No need to reimplement reference to the format or anything like that.

genet/use/recipes.py
brynpickering, line 1: I don't really understand the difference between `use` and `util`. I would have preference for "helper" functions to be in `util`

genet/use/recipes.py
```diff
@@ -0,0 +1,889 @@
...
+import pandas as pd
+from pyproj import CRS
+
+import genet as gn
+import genet.output.sanitiser as sanitiser
+from genet.utils.persistence import ensure_dir
+from genet.variables import EPSG4326
+
+
+def _to_json(dict_to_save: dict, filepath: Path) -> None:
```
brynpickering, line 18: This could(/should?) be in `output` or an `io.py` util file

genet/use/recipes.py
```diff
@@ -0,0 +1,889 @@
...
+        json.dump(sanitiser.sanitise_dictionary(dict_to_save), f, ensure_ascii=False, indent=4)
+
+
+def _write_scaled_vehicles(schedule, list_of_scales, output_dir):
+    for i in list_of_scales:
+        scale = float(i) / 100
+        schedule.scale_vehicle_capacity(scale, scale, output_dir)
+
+
+def _generate_validation_reports(network, output_dir: Path) -> None:
```
brynpickering, line 29: This could also be in an `io` util (anything related to input/output)

genet/use/recipes.py
```diff
@@ -0,0 +1,889 @@
...
+    return network
+
+
+def _cast_catchment(network_spatial_tree, df_stops, distance):
+    return network_spatial_tree.closest_links(
+        gdf_points=df_stops, distance_radius=distance
+    ).dropna()
+
+
+def _find_closest_links_by_step(
```
brynpickering, line 88: This could be in an existing util file (`spatial.py`? `graph_operations.py`?)

genet/use/recipes.py
```diff
@@ -0,0 +1,889 @@
...
+                df_stops["id"].isin(stop_ids - set(nodes.index)), ["id", "geometry"]
+            ].copy(),
+            distance=distance,
+        )
+        _df["catchment"] = distance
+        nodes = pd.concat([nodes, _df])
+    return nodes
+
+
+def _generate_modal_network_geojsons(network, modes, output_dir, filename_suffix):
```
brynpickering, line 125: `output/geojson.py`?

genet/use/recipes.py
```diff
@@ -0,0 +1,889 @@
...
+    logging.info(f"Generating visual outputs {filename_suffix}")
+    gdf = network.to_geodataframe()["links"].to_crs(EPSG4326)
+    for mode in modes:
+        _gdf = gdf[gdf["modes"].apply(lambda x: mode in x)]
+        _gdf["modes"] = _gdf["modes"].apply(lambda x: ",".join(sorted(list(x))))
+        gn.geojson.save_geodataframe(_gdf, f"mode_{mode}_{filename_suffix}", output_dir)
+
+
+def add_elevation_to_network(
```
brynpickering, line 134: `utils/elevation.py`?

genet/use/recipes.py
```diff
@@ -0,0 +1,889 @@
...
+        df = df[["id", "slope"]]
+        gdf_links = pd.merge(gdf, df, on="id")
+        gn.geojson.save_geodataframe(gdf_links.to_crs(EPSG4326), "link_slope", supporting_outputs)
+    if write_slope_to_object_attribute_file:
+        gn.elevation.write_slope_xml(slope_dictionary, output_dir)
+    logging.info("Writing the updated network")
+    network.write_to_matsim(output_dir)
+
+
+def auto_schedule_fixes(
```
brynpickering, line 202: `modify/schedule.py`?

genet/use/recipes.py
brynpickering: e.g., `output/json.py` Doesn't matter if it's just one method I'd say

tests/test_cli.py
brynpickering, line 1: If you stick with this approach, then you should parametrize all the functions into one test function since the only difference between them is the string being passed to the factory class.

brynpickering (changes requested):
It does make sense to move the helper functions out of the CLI file but it would be good to have some of them publicly accessible in an obvious place (`use/recipes.py` is not an obvious place to look for anything).

I'm also not convinced by the changes made to the test structure. The old structure might have had some repetition, but it benefitted from being quite clear. The new implementation is quite convoluted and may make maintenance quite difficult. 

It's not worth going down the rabbit hole of checking every function call's passed args to sub-functions. If doing it for CLI funcs, why not every single other function in the code? I would rather just unit test the helper functions and then integration test the CLI (if you're very worried then open some of the generated files and confirm that they include some expected string snippets or rows of data).

I won't block this PR if the test structure stays as-is. Just a note to think about maintainability not only from the perspective of code duplication but also from new users coming in and needing to add to / modify what you've written.

genet/use/recipes.py
mfitz: +1 for `util` over `use`

genet/use/recipes.py
mfitz: I like the idea of moving IO functions into an `io` module. It's a natural separation.

PR #205, opened 2023-11-03
Fix test fixtures

brynpickering:
Restructure how test fixtures are set up to align with `pytest` recommendations. 

I can't get some of the pyproj-related tests working on my machine, hence why there are several failing commits before I manage to fix everything properly. I would suggest merging this PR with "squash and merge" to mitigate the dirty commit history on `main`.

Summary:

* Use the automatically recognised `conftest.py` file to store all fixtures / utility functions , whenever they are shared by multiple test modules.
* Remove direct importing of fixtures/utils in place of implicit avaiability by virtue of them being in `conftest.py`.
* Keep only those fixtures used solely in one test module in those specific test modules.
* Add a test for `replace_link_on_pt_route` and fix that method for the tests to pass (would be worth checking that I was testing actual expected functionality @KasiaKoz!).

The benefit of utils being pytest fixtures is that you know that they are test-related by the fact that the test functions have them in their argument list. It also allows you to remove the test directory from the sys path (by removing __init__.py), which is a cleaner way to manage test dirs.


KasiaKoz (approved):
nice one, just a small comment on naming

mfitz (approved):

PR #207, opened 2023-11-08
fix a switcheroo in summary method

KasiaKoz:


tests/test_core_network.py
```diff
@@ -5686,7 +5686,10 @@ def test_generating_summary_report(network_for_summary_stats):
     report = network_for_summary_stats.summary_report()
     correct_report = {
         "network": {
-            "network_graph_info": {"number_of_links": 2, "number_of_nodes": 3},
+            "network_graph_info": {
+                "number_of_links": len(list(network_for_summary_stats.links())),
```
mfitz, line 5690: I like the replacement of implicit knowledge here - definitely nicer! 👍

mfitz (approved):

PR #208, opened 2023-11-27
Move from setup.py to pyproject.toml

brynpickering:
Another part of #188.

Here I've got rid of the bash scripts once and for all as they are now covered by pytest and its plugins. These are invoked in the base set of pytest args in the `pyproject.toml` file.

`genet` as a package is now `cml-genet`, even though the _module_ is still `genet`. This matches what's going on over at [pam](https://github.com/arup-group/pam) and will avoid clashes when uploading to be indexed on PyPI.

### DIFF
--- a/.github/workflows/build_pipeline.yml
+++ b/.github/workflows/build_pipeline.yml
@@ -31,6 +31,9 @@ jobs:
         pip install GDAL==3.4.1
         pip install -e '.[dev]'
 
+    - name: Install jupyter kernel
+      run: python -m ipykernel install --user --name genet
+
     - name: Unit tests
       run: pytest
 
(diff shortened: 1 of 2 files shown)


KasiaKoz (commented):
Looks fine, but I'm confused why the number of tests is different in the build, compared to the daily build on the main branch

[this branch](https://github.com/arup-group/genet/actions/runs/7017042975/job/19089517854):
```
Required test coverage of 94.0% reached. Total coverage: 94.17%
...
=========== 1088 passed, 1 xfailed, 55 warnings in 379.69s (0:06:19) ===========
```

[main](https://github.com/arup-group/genet/actions/runs/7061417597/job/19223011212):
```
Required test coverage of 94.0% reached. Total coverage: 94.01%
=========== 1067 passed, 1 xfailed, 58 warnings in 345.12s (0:05:45) ===========
```

any ideas?

Issue #42, opened 2020-12-22
Import/export to GMNS


pedrocamargo:
The General Modeling Network Specification is starting to make strides into becoming the standard network exchange format, and being able to import/export networks from that format will likely result in being able to reach all commercial packages as they include GMNS capabilities (as they did for OMX). 

https://github.com/zephyr-data-specs/GMNS

PR #210, opened 2023-12-05
Clean up example notebooks 

brynpickering:
Relatively minor change to align genet with the CML [Python template](https://github.com/arup-group/cookiecutter-pypackage/). 

- Notebooks are renamed to be machine-friendly
- Example data is moved to a sub-dir within `examples` and outputs from notebooks are confined to `examples/example_data/outputs`
- Some data is moved to the test dir where it is _only_ referenced in the test dir
- Removed unused files (one data file that is an example notebook output and the notebook smoke tests)


brynpickering:
This is blocking #214 btw @KasiaKoz

examples/1_intro_to_genet.ipynb
```diff
@@ -14,7 +14,7 @@
     "\n",
```
KasiaKoz, line 14: Below it looks like you're adding section numbers to the notebooks, I think you missed this one

examples/3_2_writing_data_GTFS.ipynb
```diff
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Writing GTFS data\n",
+    "# 3.2. Writing data: GTFS\n",
```
KasiaKoz, line 7: I found it weird that we skipped a section when numbering, I wonder if it's because we were a bit too ambitious and were planning to add functionality to write to a different model format soon https://github.com/arup-group/genet/issues/42

KasiaKoz (changes requested):
There is a script left in the notebooks folder that hasn't been addressed: https://github.com/arup-group/genet/tree/clean-example-notebooks/notebooks
I think that this folder should disappear altogether since it will be confusing to have both examples and notebooks folders. The script that's left generates the wiki pages based on notebook content. I would like to keep it, people have given a lot of positive feedback for the wiki; it also makes it easy to link people to specific sections.

On the side, what's the reason behind mix of cases in the naming of notebooks? they are partially lower case

examples/3_2_writing_data_GTFS.ipynb
brynpickering: Haha, yeah. Shall I renumber to remove this oddity?

brynpickering:
RE wiki generator script, good catch. I'll move that over to the "examples" dir for this PR. once we have GitHub pages I would remove it and the wiki in favour of the notebooks being rendered there (as is done in PAM e.g.: https://arup-group.github.io/pam/v0.3.1/examples/01_basics/).

examples/3_2_writing_data_GTFS.ipynb
KasiaKoz: yeah i think what you did here with renumbering is good

KasiaKoz (changes requested):
Most of the jupyter notebooks are duplicated, something must have gone wrong with renaming them?

brynpickering:
Damn... Git is generally case insensitive so I had to override a config option to make it see the case changes. I don't see duplicates locally so clearly this config change made it very confused.

brynpickering:
Force pushed a change. `git mv upper_case_file lower_case_file` is our friend.

KasiaKoz (approved):
good to go!

Issue #213, opened 2023-12-07
PROJ, proj-data, and accurate reprojections


brynpickering:
I have been wondering for a while why my local build fails a bunch of tests every time and it turns out that there is a difference in the contents of the `PROJ` library on [conda-forge](https://anaconda.org/conda-forge/proj) and on e.g. [Homebrew](https://formulae.brew.sh/formula/proj). 

The conda-forge version sets an environment variable that leads to transformation gridfiles being pulled from the internet where possible; the homebrew version doesn't set this but downloads the gridfiles and stores them locally on installation (see [here](https://github.com/Homebrew/homebrew-core/blob/fa72b1b3bc990d46a673023d2bd308e0c62be3c7/Formula/p/proj.rb#L32-L34)). To get these same data files, the conda-forge [proj-data](https://anaconda.org/conda-forge/proj-data/files) library also needs to be installed. It's a big boy: ~700MB.

The conda-forge PROJ has the advantage of a lower footprint on disk, but the attempt to query online gridfiles has an associated bug which I've come across (and which led me down this rabbit hole): https://github.com/pyproj4/pyproj/issues/705

Oddly enough, genet tests only pass with conda packages installed if I do _not_ install `proj-data` _and_ force PROJ to use its internal (quite simple) transformation grids (`pyproj.network.set_network_enabled(False)` in `genet.__init__.py`). This suggests that we are not accessing this large transformation library by default. If our tests are running on the simplified transformations, there are possible accuracy improvements to be had...

brynpickering:
In addition, I've checked the Docker image builds on AWS (do a search for `libproj` / `proj-data`) and the use of `apt-get` is leading to the _old_ transformations being installed (PROJ v7.x) which we found led to different results than the more current versions of PROJ (>= v8.x). We discussed that issue in #167.

On the GitHub CI it is v8.x being installed. 

Setting minimum versions with apt-get seems to be a pain, so this is another reason to move to a fully conda-fied install/build pipeline where we can be more explicit with version setting of PROJ and we decide if we want to include `proj-data` in an install.

PR #214, opened 2023-12-08
Update install to rely on mamba / conda envs only

brynpickering:
Somewhat supersedes #212, but is in line with the steps that #188 was broken into.

@KasiaKoz the CBC issue is solved by explictly including it in the creation of the mamba environment. I'm chasing up the CBC devs to get the windows version sorted but until then it has to be added explicitly in the install instruction (no different to now tbh) and CI tests on windows will (once I push the fix) skip tests that require CBC.


brynpickering:
Runs on Windows will only start passing once #210 is merged in (to clean up notebook names)

genet/__init__.py
```diff
@@ -19,3 +21,5 @@
 from genet.schedule_elements import Route, Schedule, Service, Stop
 from genet.use.road_pricing import Toll
 from genet.utils import elevation, google_directions, graph_operations
+
+pyproj.network.set_network_enabled(False)
```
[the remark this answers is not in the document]
brynpickering: It stops it from trying to connect to the network to download transformation gridfiles on-the-fly (see #213). If it does try, there is a bug with handling self-signed SSL certificates that will cause an error.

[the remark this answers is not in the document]
brynpickering: This wasn't needed before because the PROJ environment variable associated with this is set to OFF when downloading from homebrew/apt, but is ON when downloading from conda-forge (because in the former it automatically downloads the `proj-data` dataset too while it isn't included in the latter)

.github/workflows/pr-ci.yml
```diff
@@ -0,0 +1,48 @@
+name: Pull Request CI
+
+on:
+  pull_request:
```
KasiaKoz, line 4: will this run for a draft PR too?

README.md
```diff
@@ -114,103 +108,71 @@ as part of your command:
...
+### Installing a development environment
+
+To create a development environment for genet, with all libraries required for development and quality assurance installed, it is easiest to install genet using the [mamba](https://mamba.readthedocs.io/en/latest/index.html) package manager, as follows:
+
+1. Install mamba with the [Mambaforge](https://github.com/conda-forge/miniforge#mambaforge) executable for your operating system.
+2. Open the command line (or the "miniforge prompt" in Windows).
+3. Download (a.k.a., clone) the genet repository: `git clone git@github.com:arup-group/genet.git`
+4. Change into the `genet` directory: `cd genet`
+5. Create the genet mamba environment: `mamba create -n genet -c conda-forge -c city-modelling-lab --file requirements/base.txt --file requirements/dev.txt`
+6. Activate the genet mamba environment: `mamba activate genet`
```
KasiaKoz, line 129: I've gone through the install steps, all worked fine except here i needed to use `micromamba activate ..` command instead as `mamba activate ..` did not find the environment (I used `mamba create ..` to create it and it worked fine though..)

KasiaKoz (approved):

.github/workflows/pr-ci.yml
brynpickering: yes, I believe it will

README.md
brynpickering: sounds like you've initialised `micromamba` in your shell on your device (`micromamba init [name-of-shell-of-choice]`) but not `mamba` (similarly, `mamba init [name-of-shell-of-choice]`)

brynpickering:
@mfitz RE image size. It is smaller than the old image (incl. NodeJS) by about 5% but larger than the image in #212 by about 6%.

brynpickering:
@KasiaKoz I had to make another change for things to work on window: remove `:` from filenames. This affects one of GeNet's outputs: `vph_all_modes_within_{h - 1}:30-{h}:30.geojson`. I opted for `_` to replace `:`.

KasiaKoz (approved):

Issue #156, opened 2022-11-03
Links geojsons by OSM tag in standard outputs


ana-kop:
Standard outputs script is supposed to generate geojsons of network links by OSM tag. However, this has not been happening since the structure of link attributes has been changed from

```
highway_tags = n.link_attribute_data_under_key({'attributes': {'osm:way:highway': 'text'}})
```

to

```
highway_tags = n.link_attribute_data_under_key({'attributes': 'osm:way:highway'})
``` 

Need to add functionality to ensure that geojsons get made both for the old and the new link attribute structure.

PR #217, opened 2023-12-19
Make geoparquet the new standard for ouptuts

KasiaKoz:
GeNet generates a selection of geojson outputs for network eye-ball validation and visualisation.
For large networks these can be very slow and clunky. The files also take up more space on disk.
So with @Theodore-Chatziioannou suggestion, this PR changes the default output format for standard outputs. It also tidies up the three spatial output file options for networks/schedules.

Londinium network links output goes from 13.8MB to 2.2MB 🚀 

Since I was already in that neck of the woods (standard outputs generation), I stumbled on a known problem and fixed: https://github.com/arup-group/genet/issues/156

### DIFF
--- a/genet/core.py
+++ b/genet/core.py
@@ -2613,7 +2613,7 @@ class Network:
         self, output_dir, gtfs_day="19700101", include_geojson_files=False, include_shp_files=False
     ):
         """
-        Generates spatial geoparquet that can be used for generating standard visualisations.
+        Generates spatial geoparquet files that can be used for visualisations.
         These can also be used for validating network for example inspecting link capacity, freespeed, number of lanes,
         the shape of modal subgraphs.
         :param output_dir: path to folder where to save resulting files
--- a/genet/core.py
+++ b/genet/core.py
@@ -2609,17 +2609,27 @@ class Network:
             )
         return con_desc
 
-    def generate_standard_outputs(self, output_dir, gtfs_day="19700101", include_shp_files=False):
+    def generate_standard_outputs(
+        self, output_dir, gtfs_day="19700101", include_geojson_files=False, include_shp_files=False
+    ):
         """
-        Generates geojsons that can be used for generating standard kepler visualisations.
+        Generates spatial geoparquet that can be used for generating standard visualisations.
         These can also be used for validating network for example inspecting link capacity, freespeed, number of lanes,
         the shape of modal subgraphs.
-        :param output_dir: path to folder where to save resulting geojsons
+        :param output_dir: path to folder where to save resulting files
         :param gtfs_day: day in format YYYYMMDD for the network's schedule for consistency in visualisations,
         defaults to 1970/01/01 otherwise
+        :param include_geojson_files: whether to generate geojson outputs as well
+        :param include_shp_files: whether to generate shape file outputs as well
         :return: None
         """
-        geojson.generate_standard_outputs(self, output_dir, gtfs_day, include_shp_files)
+        geojson.generate_standard_outputs(
+            self,
+            output_dir,
+            gtfs_day=gtfs_day,
+            include_shp_files=include_shp_files,
+            include_geojson_files=include_geojson_files,
+        )
         logging.info("Finished generating standard outputs. Zipping folder.")
         persistence.zip_folder(output_dir)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,11 +3,13 @@
 ## [Unreleased]
 
 ### Fixed
+* Fixed generating standard outputs by highway tag which were broken after moving to to storing additional attributes in short form [#217](https://github.com/arup-group/genet/pull/217)
```
brynpickering, line 6: `to to`

genet/core.py
```diff
@@ -2609,17 +2609,27 @@ def check_connectivity_for_mode(self, mode):
             )
         return con_desc
 
-    def generate_standard_outputs(self, output_dir, gtfs_day="19700101", include_shp_files=False):
+    def generate_standard_outputs(
```
brynpickering, line 2612: Why is it an option to produce geojson/shapefile in _addition_ to geoparquet rather than _instead of_?

genet/core.py
```diff
@@ -2697,31 +2707,78 @@ def write_to_json(self, output_dir):
             self.schedule.write_to_json(output_dir)
         self.write_extras(output_dir)
 
-    def write_to_geojson(self, output_dir, epsg: str = None):
+    def write_spatial(
```
brynpickering, line 2710: If `write_spatial` is accessed from the `write_to_...` funcs lower down, maybe it should be a private function?

genet/core.py
```diff
@@ -2697,31 +2707,78 @@ def write_to_json(self, output_dir):
...
+                to_shp=to_shp,
+            )
         if self.schedule:
-            self.schedule.write_to_geojson(output_dir, epsg)
+            self.schedule.write_spatial(
+                output_dir, epsg=epsg, to_parquet=to_parquet, to_geojson=to_geojson, to_shp=to_shp
+            )
         self.write_extras(output_dir)
 
+    def write_to_geojson(self, output_dir, epsg: str = None):
```
brynpickering, line 2755: maybe just `to_geojson`? This would align with naming conventions in other libraries (`pandas`, `geopandas`, `xarray`, ...)

genet/output/geojson.py
brynpickering, line 1: Maybe the filename needs to change? Or `save_geodataframe` should be in something like an `io.py` file.

genet/output/geojson.py
```diff
@@ -59,28 +59,53 @@ def line_geometry():
     return {"nodes": nodes, "links": links}
 
 
-def save_geodataframe(gdf, filename, output_dir, include_shp_files=False):
+def save_geodataframe(
```
brynpickering, line 62: There's some repetition here with `write_spatial` in `core.py`. Not sure where it is best placed. Maybe `write_spatial` should defer the check for whether one of `to_...` is defined to `save_geodataframe`

genet/output/sanitiser.py
```diff
@@ -14,11 +14,10 @@ def sanitise_list(x):
 def sanitise_geodataframe(gdf):
     if isinstance(gdf, GeoSeries):
         gdf = GeoDataFrame(gdf)
-    gdf = gdf.fillna("None")
     object_columns = gdf.select_dtypes(["object"]).columns
     for col in object_columns:
         if gdf[col].apply(lambda x: isinstance(x, (set, list))).any():
-            gdf[col] = gdf[col].apply(lambda x: ",".join(x))
+            gdf[col] = gdf[col].apply(lambda x: ",".join(x) if isinstance(x, (list, set)) else x)
```
brynpickering, line 20: How is this handled when loading in data into genet that has previously gone through genet's sanitiser?

genet/schedule_elements.py
```diff
@@ -3587,29 +3595,74 @@ def write_to_json(self, output_dir):
...
-        :param epsg: projection if the geometry is to be reprojected, defaults to own projection
-        :return:
-        """
+        :param epsg: projection, if the geometry is to be reprojected, defaults to own projection
+        :param to_parquet: whether to save to parquet file format, defaults to no
+        :param to_geojson: whether to save to geojson file format, defaults to no
+        :param to_shp: whether to save to shape file format, defaults to no
+        :return: None, files are saved to disk
+        """
+        if not any((to_parquet, to_geojson, to_shp)):
```
brynpickering, line 3611: Again, I'd defer this to `save_geodataframe` so that you dont' have the repetition.

requirements/base.txt
```diff
@@ -15,6 +15,7 @@ osmnx < 2
 pandas >= 1.5, < 2.1
 polyline >= 2, < 3
 pre-commit < 4
+pyarrow > 14
```
brynpickering, line 18: Just to ensure that pyarrow can work with GDAL < 3.6 (something windows users might want to force, see: https://github.com/OSGeo/gdal/issues/8959) I would set the minimum to v12. Then cap it to v15, so we don't get caught out by breaking changes that could come in unexpectedly

brynpickering (changes requested):
Looks like a worthwhile change. I'm unsure what the benefits are of being able to dump to multiple file formats simultaneously. If there is a reason you've kept this functionality then all good. If there is no obvious use-case then I would simplify the code to having a `to_file` method that infers the driver to use based on the file extension.

Maybe the README should have something on geoparquet files (and point to some online source that discusses how to use them with e.g. QGIS?)

genet/core.py
KasiaKoz: It started with just geojson, and then some transport people we were working with wanted shapefiles. Because we still preferred geojson, we just wanted an option to also output shapefiles if necessary. But I agree, it's better to have a single option and just call on a method twice with different options if you need other types

genet/core.py
KasiaKoz: got rid of the the `write_to_...` methods

genet/output/geojson.py
KasiaKoz: made an `io` module

genet/output/geojson.py
KasiaKoz: made a check method that gets called when saving but reuse it in core too (because it takes a bit of time to format the network, so you don't want to wait a while to find out the format you want is not right)

genet/output/sanitiser.py
KasiaKoz: we're not reading networks from geojson/etc files

src/genet/schedule_elements.py
```diff
@@ -3697,28 +3698,38 @@ def write_to_json(self, output_dir: str):
             json.dump(self.to_json(), outfile)
         self.write_extras(output_dir)
 
-    def write_to_geojson(self, output_dir: str, epsg: str):
-        """Writes Schedule graph to nodes and edges geojson files.
+    def write_spatial(self, output_dir, epsg: str = None, filetype: Optional[str] = "parquet"):
```
brynpickering, line 3701:
```suggestion
    def write_spatial(self, output_dir, epsg: str = None, filetype: str = "parquet"):
```

`Optional` is used when you allow the value to be `None` (i.e., it's equivalent to `str | None`).

src/genet/utils/io.py
```diff
@@ -0,0 +1,56 @@
+import os
+from typing import Optional
+
+import geopandas as gpd
+from pandas.core.dtypes.common import is_datetime64_any_dtype as is_datetime
+
+from genet.output import sanitiser as sanitiser
+from genet.utils import persistence as persistence
```
brynpickering, line 8:
```suggestion
from genet.output import sanitiser
from genet.utils import persistence
```

src/genet/utils/io.py
```diff
@@ -0,0 +1,56 @@
...
+        RuntimeError: `filetype` is not a supported file type
+    """
+    if filetype.lower() not in SUPPORTED_FILE_FORMATS:
+        raise RuntimeError(
+            f"{filetype} is not a supported file type: {', '.join(SUPPORTED_FILE_FORMATS)}"
+        )
+
+
+def save_geodataframe(
+    gdf: gpd.GeoDataFrame, filename: str, output_dir: str, filetype: Optional[str] = "parquet"
```
brynpickering, line 29:
```suggestion
    gdf: gpd.GeoDataFrame, filename: str, output_dir: str, filetype: str = "parquet"
```

tests/test_utils_io.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def test_throws_error_when_filetype_is_not_supported():
+    with pytest.raises(RuntimeError) as e:
+        genet.utils.io.check_file_type_is_supported("hello")
+
+    assert "is not a supported file type" in str(e.value)
+
+
+def test_saving_geodataframe_with_missing_geometry_produces_file(tmpdir):
```
brynpickering, line 16: Maybe add another test that checks the expected behaviour if the gdf is completely empty (i.e. no file exists)?

brynpickering (approved):
Just a few minor comments / suggestions, otherwise good to go I think!

PR #225, opened 2024-01-25
Fix references to stop names changing during snapping+routing

### LINKED ISSUE #224: Reference to stop name changes during snapping+routing on network
The step to [process outputs of pyomo model](https://github.com/arup-group/genet/blob/main/genet/max_stable_set.py#L301) strips too much of the string. It changes the name of the stop, if the stop name begins with an 'x' - this leads to routing issues because the resulting stop doesn't actually exist in the network.

```python

2024-01-25 11:32:24,827 - Passing problem to solver
--
2024-01-25 11:32:24,854 -
Routing Service: `some_route_id` resulted in the following Exception:
Traceback (most recent call last):
File "/src/genet/core.py", line 1827, in route_schedule
mss = modify_schedule.route_pt_graph(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/src/genet/modify/schedule.py", line 64, in route_pt_graph
mss.solve(solver=solver)
File "/src/genet/max_stable_set.py", line 306, in solve
self.solution = {
^
File "/src/genet/max_stable_set.py", line 307, in <dictcomp>
self.problem_graph.nodes[node]["id"]: self.problem_graph.nodes[node]["link_id"]
~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
File "/usr/local/lib/python3.11/site-packages/networkx/classes/reportviews.py", line 194, in __getitem__
return self._nodes[n]
~~~~~~~~~~~^^^
KeyError: '_stop.link:1234'
```
with some stop originally named `x_stop`

KasiaKoz:
Fixes #224 

@brynpickering - maybe you know of a better way of pulling results out from a pyomo model? (Bearing in mind I want to test this without having to mock loads of objects...)

### DIFF
--- a/genet/max_stable_set.py
+++ b/genet/max_stable_set.py
@@ -21,7 +21,7 @@ def has_attrib(attrib_value):
 
 
 def process_model_result(result_list: List[str]):
-    return [v.strip("x[\\']") for v in result_list]
+    return [v.replace("x['", "").replace("']", "") for v in result_list]
 
 
 class MaxStableSet:
--- a/genet/max_stable_set.py
+++ b/genet/max_stable_set.py
@@ -1,7 +1,6 @@
 import itertools
 import logging
 from copy import deepcopy
-from typing import List
 
 import matplotlib.pyplot as plt
 import networkx as nx
@@ -20,8 +19,12 @@ def has_attrib(attrib_value):
     return True
 
 
-def process_model_result(result_list: List[str]):
-    return [v.replace("x['", "").replace("']", "") for v in result_list]
+def get_indices_of_chosen_problem_graph_nodes(model: pe.Model):
+    return [
+        v.index()
+        for v in model.component_data_objects(pe.Var)
+        if v.value is not None and float(v.value) == 1.0
+    ]
 
 
 class MaxStableSet:
@@ -302,13 +305,8 @@ class MaxStableSet:
             # Solution parse
             # --------------------------------------------------------
 
-            selected = process_model_result(
-                [
-                    str(v)
-                    for v in model.component_data_objects(pe.Var)
-                    if v.value is not None and float(v.value) == 1.0
-                ]
-            )
+            selected = get_indices_of_chosen_problem_graph_nodes(model)
+
             # solution maps Stop IDs to Link IDs
             self.solution = {
                 self.problem_graph.nodes[node]["id"]: self.problem_graph.nodes[node]["link_id"]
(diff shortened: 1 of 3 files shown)


brynpickering:
How about `pd.Series(variable.extract_values())`? that's [what we use in calliope](https://github.com/calliope-project/calliope/blob/aad664ff1202d298e3265cd8994ca5e9a57788e9/calliope/backend/pyomo/util.py#L141)

genet/max_stable_set.py
```diff
@@ -19,6 +20,10 @@ def has_attrib(attrib_value):
     return True
 
 
+def process_model_result(result_list: List[str]):
```
mfitz, line 23: I am not sure about the name of this function. Is it processing model results? From what I can see, it's transforming stop names. Or is it more generic than that? Can we come up with a more descriptive name?

tests/test_max_stable_set.py
```diff
@@ -1522,3 +1523,22 @@ def test_combining_two_changesets_with_overlap(assert_semantically_equal, partia
...
+            ["stop_3.link:link_7_8_car", "stop_1.link:link_4_5_car"],
+        ),
+        (
+            "stops with x in name",
+            ["x['x_stop.link:1234']", "x['stopx.link:1234']", "x['stxop.link:1234']"],
+            ["x_stop.link:1234", "stopx.link:1234", "stxop.link:1234"],
+        ),
+    ],
+)
+def test_processing_model_results(case_name, model_output, expected):
```
mfitz, line 1543: Great to see a dedicated unit test for this 👍 

As with the function under test though, I am not sure about the name.  How would you describe what we expect this new function to do? The current test name doesn't give me any clue about the behaviour we're verifying here. Is there a way to work the expected behaviour into the test name in a `test_does_something_when_something` style?

mfitz (changes requested):
The fix looks good, but I think we can come up with names for the new function and its test that give a better description of the behaviour.

KasiaKoz:
I pulled out the entire code block that processes the model for solution variables - I didn't want to do it initially, because I really didn't want to mock loads of objects, but it turned out to not be that bad at all (no need to even use mocker). I noticed there was a dedicated method to get the index of the variable, which is much cleaner than processing the string representation of the variable.
I hope the method and test names look a lot better now too - let me know what you think @mfitz !

mfitz (approved):
Looks good! 👍

PR #231, opened 2024-03-08
Rename Route.route attribute

### LINKED ISSUE #228: BUG: `route` is both a method and an attribute of `Route`
`Route.route` is both a list attribute and a method that takes a route ID returns `self` if the input route ID == `self.id`.

brynpickering:
Fixes #228 

I've renamed the `Route.route` attribute to `Route.network_route`, but as mentioned in the issue, I think it might be better named as, e.g.:

* `Route.links`
* `Route.network_links`
* `Route.route_links`
* `Route.network`

`route_links` seems unnecessarily verbose. You know it's an attribute of a `Route`, so why prepend with `route_`... By extension, `route_short_name` and `route_long_name` could simply be `short_name` and `long_name`.

Anyway, now that I've made the main name change, it's a simple job of a repo-wide find and replace to update `network_route` to the attribute name we finally settle on.


KasiaKoz:
I think `Route.network_links` works best - it tells you that it's referring to links in a network, that there is a link here between the two, and a network is a whole other object, so it might prevent people looking for links where they don't exist..

brynpickering:
Updated to `network_links`. I noticed some overlap with the use of that name for variables in plotting and for a particular output file. Not really an issue except that global find/replace will pick those up too.

KasiaKoz (changes requested):
Looks great, one thing missing and then it'll be ready to go - changelog needs an update with a highlight that it's a breaking change

brynpickering:
> changelog needs an update with a highlight that it's a breaking change

Good point. Updated!

KasiaKoz (approved):
✨

PR #246, opened 2024-11-07
Convenience method to strip links of mode

### LINKED ISSUE #243: Method to strip links of mode for core `Network` object
This method requires to specify links from which mode `X` should be removed:

https://github.com/arup-group/genet/blob/afa6e0f2219f8e1069fe56667e4d6517210780ab/src/genet/core.py#L746

It would be useful to have a default that strips mode `X` from all links.

KasiaKoz:
Closes #243 

I lowered code coverage because I have no idea why my additions would have lowered it, I added only a couple of lines and tested them...

### DIFF
--- a/tests/test_core_network.py
+++ b/tests/test_core_network.py
@@ -2342,8 +2342,8 @@ def test_removing_mode_from_all_links_retains_other_modes_on_links():
 
     n.remove_mode_from_all_links(mode="bike")
 
-    assert n.link("0")["modes"] == {"car", "piggyback"}
-    assert n.link("1")["modes"] == {"car"}
+    assert n.link("0")["modes"] == {"car", "piggyback"}, "The link modes have been incorrectly set"
+    assert n.link("1")["modes"] == {"car"}, "The link modes have been incorrectly set"
 
 
 def test_removing_mode_from_all_links_deletes_links_with_no_other_modes():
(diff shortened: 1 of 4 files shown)


tests/test_core_network.py
```diff
@@ -2335,6 +2335,40 @@ def test_subnetwork_on_spatial_condition_delagates_to_spatial_methods_to_get_sub
...
+    n.remove_mode_from_all_links(mode="bike")
+
+    assert n.link("0")["modes"] == {"car", "piggyback"}, "The link modes have been incorrectly set"
+    assert n.link("1")["modes"] == {"car"}, "The link modes have been incorrectly set"
+
+
+def test_removing_mode_from_all_links_deletes_links_with_no_other_modes():
+    n = Network("epsg:27700")
+    n.add_link("0", 1, 2, attribs={"modes": {"car", "bike"}, "length": 1})
+    n.add_link("1", 2, 3, attribs={"modes": {"bike"}, "length": 1})
```
mfitz, line 2352: This is not directly about this PR, but this test raises a question: Is there some warning in the logs or elsewhere when you remove all modes from a link and that link is subsequently removed from the graph? I get that a link with no modes cannot be traversed, and so can be removed, but the user might have done that inadvertently, and the downstream effects could be very confusing if so.

mfitz (approved):
Looks good. I have asked a question prompted by one of the new tests, but it's not about anything introduced in this PR.

tests/test_core_network.py
KasiaKoz: There are logs when links are removed. The removal of links is not automatic, unless you use these dedicated methods. So a user could mess around with modes on the links and the structure of the graph remains the same. In either case, whether the links are removed or not (so, just modes missing from links) genet has methods to verify validity of the graph and if they pose issues these will come out in the reports. These run automatically if using genet cli commands, or can/should be called upon if a user is using genet as a package and expecting to use their network with matsim.

Issue #244, opened 2024-11-05
Method to split links of mode for core `Network` object


KasiaKoz:
In similar vein to https://github.com/arup-group/genet/issues/243. It would also be useful to have a 'split' method for modes, right now we have a cli command that does this:

https://github.com/arup-group/genet/blob/afa6e0f2219f8e1069fe56667e4d6517210780ab/src/genet/cli.py#L1027

but a method of the object would make it easily reusable in other circumstances

PR #247, opened 2024-11-12
Method to split links depending on mode

KasiaKoz:
Closes: #244, second in a series of 3, to facilitate better active mode support

### DIFF
--- a/src/genet/core.py
+++ b/src/genet/core.py
@@ -785,16 +785,33 @@ class Network:
         self.remove_links(no_mode_links)
 
     def split_links_on_mode(self, mode: str, link_id_prefix: Optional[str] = None) -> Set[str]:
-        if link_id_prefix is None:
-            link_id_prefix = f"{mode}---"
+        """Method to split links depending on mode.
+        Existing links with mode `mode` will have that mode removed.
+        New links will be added with only the mode `mode` and inheriting data from the link they originated from.
+        The IDs of ne link IDs will by default identify the mode, but can be changed with `link_id_prefix`.
+
+        Args:
+            mode (str): Mode to split from the links.
+            link_id_prefix (str): Optional, you can request what the
+
+        Returns:
+            Set of link IDs of the new links
+        """
         modal_links = self.links_on_modal_condition({mode})
+        modal_links = list(modal_links)
+
+        if link_id_prefix == "":
+            logging.warning("Empty string was set as prefix, the IDs will be randomly assigned")
+            new_link_ids = self.generate_indices_for_n_edges(len(modal_links))
+        else:
+            if link_id_prefix is None:
+                link_id_prefix = f"{mode}---"
+            new_link_ids = [f"{link_id_prefix}{link_id}" for link_id in modal_links]
         new_links = {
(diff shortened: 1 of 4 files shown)


src/genet/core.py
```diff
@@ -784,6 +784,56 @@ def empty_modes(mode_attrib):
...
+        Existing links with mode `mode` will have that mode removed.
+        New links will be added with only the mode `mode` and inheriting data from the link they originated from.
+        The IDs of new link IDs will by default identify the mode, but can be changed with `link_id_prefix`.
+
+        Examples:
+            ```python
+            [1] network.link("LINK_ID")
+            [out] {"id": "LINK_ID", "modes": {"car", "bike"}, "freespeed": 5, ...}
+            ```
+
```
mfitz, line 798: Am I misreading/misunderstanding this? It feels like the actual function call is missing from this example. I expected a before state, a call to the function, and an after state.

tests/test_core_network.py
```diff
@@ -2404,6 +2404,86 @@ def test_removing_mode_from_links_removes_empty_links():
...
+    ), "The new link did not inherit the same attribute value"
+
+
+def test_splitting_links_uses_desired_prefix_for_new_link_ids():
+    n = Network("epsg:27700")
+    n.add_link("0", 1, 2, attribs={"modes": {"car", "bike"}, "length": 1})
+
+    new_links = n.split_links_on_mode("bike", link_id_prefix="HEYO-")
+
+    assert len(new_links) > 0, "No new links were generated"
```
mfitz, line 2448: Will more than one new link be generated in this case?

tests/test_core_network.py
```diff
@@ -2404,6 +2404,86 @@ def test_removing_mode_from_links_removes_empty_links():
...
+
+
+def test_splitting_links_generates_unique_ids_for_new_links():
+    n = Network("epsg:27700")
+    for i in range(10):
+        n.add_link(str(i), i, i + 1, attribs={"modes": {"car", "bike"}, "length": 1})
+
+    new_links = n.split_links_on_mode("bike", link_id_prefix="HEYO-")
+
+    new_links = set(new_links)  # To force uniqueness
```
mfitz, line 2462: Isn't `new_links` already a set? You could assert that rather than making a new set.

tests/test_core_network.py
```diff
@@ -2404,6 +2404,86 @@ def test_removing_mode_from_links_removes_empty_links():
...
+    assert len(new_links) == 10, "The number of link IDs is incorrect"
+    for link_id in new_links:
+        assert (
+            n.link(link_id)["id"] == link_id
```
mfitz, line 2466: Would you also want to assert that the link IDs all start with `HEYO-`?

mfitz (approved):
Looks good. I like the unit tests - concise and clear.

I've asked a couple of questions inline, but there's nothing crucial.

tests/test_core_network.py
KasiaKoz: no just one, I'll change it to a stricter condition and get rid of the loop below, since there's just one link

Issue #248, opened 2024-11-13
`KeyError: 'from'` when adding links


KasiaKoz:
In a very specific scenario---when adding links to graph where edges were removed, opening up multi edge indices up for grabs---the following error appears:

```
        self.graph.add_edges_from(
            [
                (
>                   attribs["from"],
                    attribs["to"],
                    add_to_link_id_mapping[link]["multi_edge_idx"],
                    attribs,
                )
                for link, attribs in links_and_attributes.items()
            ]
        )
E       KeyError: 'from'

../src/genet/core.py:1345: KeyError
```

This only happens sometimes. We rely on `networkx` to generate a unique multi index, so that we can have unique edges stored on a the graph between the same pair of nodes. Sometimes the method from `networkx` we use returns an available index that is sandwiched between the indices we already use (it was been freed by an earlier operation on the graph), and in that case, our code fails and the data for that link disappears along the way, throwing the error seen above.

Issue #245, opened 2024-11-05
Add cli command to replace or add a whole subgraph of some mode


KasiaKoz:
Coming from a project requirement. It would be useful to replace a section of the network graph. In the specific project case, we are replacing/adding active modes to an existing network. We don't want to have to rebuild the entire network just to add walking and/or cycling.
Issues #243 and #244 would help to facilitate this.

PR #249, opened 2024-11-13
Bugfix: Multi-indexing fails in a very specific case

KasiaKoz:
Fixes: #248, in a wild turn of events, my projected 3 part PR series has grown to 4. I found this bug while working on #245. The fix is inversely proportional to the amount of time it took to diagnose and recreate the issue in a unit test..

I had to mock the method which gives the available multi-index for an edge in the graph because it would not give me the same behaviour as I was seeing with a larger dataset. The problem happens when this method outputs an available multi-index that is sandwiched by multi-indices already being used on the graph. In this small example it would always give me the largest index being used +1, which did not recreate the error.

Stripped from all the context, the problem was setting a dataframe with an output of a groupby (of the same dataframe) which has inherited indexing from the groupby operation. This did not match the indexing of the original dataframe, so when trying to set it back in, it resulted in loss of data.

### DIFF
--- a/tests/test_core_network.py
+++ b/tests/test_core_network.py
@@ -1938,6 +1938,26 @@ def test_adding_multiple_links_with_id_and_multi_idx_clashes(assert_semantically
     )
 
 
+def test_adding_many_links_with_id_and_multi_idx_clashes(assert_semantically_equal, mocker):
+    n = Network("epsg:27700")
+    for i in range(3):
+        n.add_link(str(i), 1, 2)
+    n.remove_link(str(1))
+    # this method will sometimes output the smallest multiindex and sometimes just the next
+    # one after the highest used one. We fix it's return value to 1, an available index
+    # inbetween which caused this issue: https://github.com/arup-group/genet/issues/248
+    # I'm mocking this method here and not at the top of this test because adding the initial links
+    # to the graph would be impacted by it
+    mocker.patch.object(nx.MultiDiGraph, "new_edge_key", return_value=1)
+    links_to_add = {f"new-{i}": {"id": f"new-{i}", "from": 1, "to": 2} for i in range(3)}
+
+    reindexing_dict, links_and_attribs = n.add_links(links_to_add)
+
+    assert (
+        len({v["multi_edge_idx"] for v in n.link_id_mapping.values()}) == 5
+    ), "Some multiindices were not unique"
+
+
 def test_adding_multiple_links_missing_some_from_nodes():
     n = Network("epsg:27700")
     with pytest.raises(RuntimeError) as error_info:
--- a/tests/test_core_network.py
+++ b/tests/test_core_network.py
@@ -1938,7 +1938,7 @@ def test_adding_multiple_links_with_id_and_multi_idx_clashes(assert_semantically
     )
 
 
-def test_adding_many_links_with_id_and_multi_idx_clashes(assert_semantically_equal, mocker):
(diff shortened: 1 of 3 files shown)


tests/test_core_network.py
```diff
@@ -1938,6 +1938,26 @@ def test_adding_multiple_links_with_id_and_multi_idx_clashes(assert_semantically
...
+    # inbetween which caused this issue: https://github.com/arup-group/genet/issues/248
+    # I'm mocking this method here and not at the top of this test because adding the initial links
+    # to the graph would be impacted by it
+    mocker.patch.object(nx.MultiDiGraph, "new_edge_key", return_value=1)
+    links_to_add = {f"new-{i}": {"id": f"new-{i}", "from": 1, "to": 2} for i in range(3)}
+
+    reindexing_dict, links_and_attribs = n.add_links(links_to_add)
+
+    assert (
+        len({v["multi_edge_idx"] for v in n.link_id_mapping.values()}) == 5
```
mfitz, line 1957: I _think_ I understand why we expect five unique links, but let me check:

We had three links, and we removed one. We then added three new links and want to ensure that we end up with the two original links plus the three new ones in the graph. The bug effectively overwrites one of the original (or new?) links due to an index clash, so one link would be lost, and we would see only four links in the graph. Is that it?

In any case, can we clarify why we expect five links in the modified graph?

tests/test_core_network.py
```diff
@@ -1938,6 +1938,26 @@ def test_adding_multiple_links_with_id_and_multi_idx_clashes(assert_semantically
     )
 
 
+def test_adding_many_links_with_id_and_multi_idx_clashes(assert_semantically_equal, mocker):
```
mfitz, line 1941: The name of the test doesn't help me understand the intent of the test.

How about: `test_avoids_destructive_index_clashes_when_adding_replacement_links`? Or... something better than that? :-)

mfitz, line 1941: I know this isn't new in this PR, but using a PyTest fixture to pass in an assertion helper method is odd.

The PyTest fixture mechanism is for... well, test fixtures. A fixture is a concept in unit tests; it's the state or general environment needed to run the test. It's part of the setup. The helper method is not part of the test's fixture; it's part of the assertion or verification.

There are other, more obvious ways to make a helper assertion method available for the test. It feels like a misuse of the fixture mechanism to use it for things that are not fixtures.

tests/test_core_network.py
KasiaKoz: > We had three links, and we removed one. We then added three new links and want to ensure that we end up with the two original links plus the three new ones in the graph.

Yup, precisely.

> The bug effectively overwrites one of the original (or new?) links due to an index clash, so one link would be lost, and we would see only four links in the graph. Is that it?

The bug threw an exception, which is great tbh, it failed fast. There were no funny links behaviour in the network at the end, and this didn't bleed into existing networks 👍

tests/test_core_network.py
KasiaKoz: yeah that's really weird, I didn't even notice it (I accidentally copied it over from another test, I'm not using it here 😅 ) I think this change happened with the update to cookie cutter template etc (I remember things got moved to the conftest file around that time) but I don't remember why or how it ended up this way

KasiaKoz:
I also added an extra assertion, which I hope makes things clearer

mfitz (approved):
👍

PR #250, opened 2024-11-25
Add cli command for attaching modal subgraphs

### LINKED ISSUE #245: Add cli command to replace or add a whole subgraph of some mode

KasiaKoz:
Closes #245

### DIFF
--- a/src/genet/cli.py
+++ b/src/genet/cli.py
@@ -1115,7 +1115,7 @@ def replace_modal_subgraph(
     modes: str,
     increase_capacity: bool,
 ):
-    """Add extracted modal subgraphs from the `subgraph network` to the main `network` (without merging links)
+    """Add extracted modal subgraphs from the `subgraph network` to the main `network` (without merging links).
 
     This creates separate modal subgraphs for the given modes.
     It replaces any mentions or links of specified modes, in the original network, and replaces them with links taken
@@ -1141,15 +1141,18 @@ def replace_modal_subgraph(
         [out] {"id": "LINK_ID", "modes": {"car"}, "some_attrib": "network", ...}
         ```
 
-        The new bike links will make their way from the subgraph network, with just the intended mode,
-        inheriting data from the subgraph links
+        The new bike links will make their way from the subgraph network, with just the intended mode.
+        The links will retain all data from the subgraph links.
+        Their ID will change to indicate the modal subgraph they belong to.
         ```python
         [1] network.link("bike---SUBNET_LINK_ID")`
         [out] {"id": "bike---SUBNET_LINK_ID", "modes": {"bike"}, "some_attrib": "sub_network", ...}
         ```
 
-        In the case when a link already has a single dedicated mode in the network, that link will be removed.
-        For other links in the network, their allowed modes may have changed.
+        Nodes related to the subgraph links will also be added, only once
+        (if multiple modes are requested, and they share nodes)
+        In the case when a link in the original network has a single dedicated mode, that link will be removed.
+        For other links in the network, their allowed modes will have changed if they allowed the requested modes.
         So, any simulation outputs may not be valid with this new network.
     """
     modes = modes.split(",")
@@ -1195,7 +1198,7 @@ def replace_modal_subgraph(
     )
     network.add_nodes(new_nodes)
 
-    # get links from each modal subgraph
+    # strip and add links for each modal subgraph
     for mode in modes:
         link_id_prefix = f"{mode}---"
 
--- a/src/genet/cli.py
+++ b/src/genet/cli.py
@@ -1242,6 +1242,11 @@ def replace_modal_subgraph(
             mode_links = {link_id: {"capacity": 9999} for link_id in modal_links}
             network.apply_attributes_to_links(mode_links)
 
+    # finally check and remove any isolated nodes that may have been left after removing modal links from the
+    # original network
+    logging.info("Checking for isolated nodes")
+    network.remove_isolated_nodes()
+
     # report on final state, save outputs
     logging.info(f"Number of nodes after adding modal graphs: {len(list(network.nodes()))}")
     logging.info(f"Number of links after adding modal graphs: {len(network.link_id_mapping)}")
--- a/src/genet/cli.py
+++ b/src/genet/cli.py
@@ -1063,12 +1063,178 @@ def separate_modes_in_network(
         logging.info(f"Splitting links for mode: {mode}")
         new_links = network.split_links_on_mode(mode)
         if increase_capacity:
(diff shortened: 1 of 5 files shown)


src/genet/cli.py
```diff
@@ -136,7 +136,7 @@ def _generate_modal_network_geojsons(network, modes, output_dir, filename_suffix
     gdf = network.to_geodataframe()["links"].to_crs(EPSG4326)
     for mode in modes:
         _gdf = gdf[gdf["modes"].apply(lambda x: mode in x)]
-        _gdf["modes"] = _gdf["modes"].apply(lambda x: ",".join(sorted(list(x))))
+        _gdf.loc[:, "modes"] = _gdf.loc[:, "modes"].apply(lambda x: ",".join(sorted(list(x))))
```
KasiaKoz, line 139: this was a standard pandas setting with copy:
```python
/Users/kasia.kozlowska/PycharmProjects/CML/genet/venv/lib/python3.12/site-packages/geopandas/geodataframe.py:1528: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  super().__setitem__(key, value)
```

src/genet/core.py
```diff
@@ -774,7 +774,8 @@ def empty_modes(mode_attrib):
 
         df["modes"] = df["modes"].apply(lambda x: persistence.setify(x))
 
-        df = df.loc[df.index.intersection(links)][df["modes"].apply(lambda x: bool(mode & x))]
+        df = df.loc[df.index.intersection(links)]
+        df = df[df["modes"].apply(lambda x: bool(mode & x))]
```
KasiaKoz, line 778: I think this one had bug potential, those operations should not have been chained this way
The warning:
```python
/Users/kasia.kozlowska/PycharmProjects/CML/genet/src/genet/core.py:777: UserWarning: Boolean Series key will be reindexed to match DataFrame index.
  df = df.loc[df.index.intersection(links)][df["modes"].apply(lambda x: bool(mode & x))]
```

tests/test_cli_replace_modal_subgraph.py
```diff
@@ -0,0 +1,195 @@
...
+        projection="epsg:27700",
+        output_dir=tmpdir,
+        path_to_subgraph=sub_network["path"],
+        modes="bike",
+        increase_capacity=True,
+    )
+
+    output_data = output_network.link(sub_network["expected_bike_link_id"])
+    keys_to_ignore = {"from", "s2_from", "to", "s2_to", "id", "modes", "capacity"}
+    assert {k: v for k, v in output_data.items() if k not in keys_to_ignore} == {
```
mfitz, line 101: Maybe this assertion can be extracted into a helper method like `assert_link_data_equal(link_1, link_2)` that would keep the intent of the assertion clear but remove some of the noisy detail of the comparison mechanics and make the test a little easier to understand.

tests/test_cli_replace_modal_subgraph.py
```diff
@@ -0,0 +1,195 @@
...
+    ), f"Link {expected_new_link_id} did not have capacity increased"
+
+    link_id = network["link_id"]
+    assert output_network.has_link(link_id), f"Link {link_id} is missing from the output network"
+    assert (
+        output_network.link(link_id)["capacity"] == original_net_capacity[link_id]
+    ), f"Link {link_id} did not retain th original capacity value"
+
+
+def test_does_not_increase_capacity_if_not_requested(tmpdir, network, sub_network):
```
mfitz, line 135: `test_preserves_link_capacity_value_when_replacing_modal_subgraphs` or suchlike, maybe?

tests/test_cli_replace_modal_subgraph.py
```diff
@@ -0,0 +1,195 @@
...
+    output_network = cli._replace_modal_subgraph(
+        path_to_network=network["path"],
+        projection="epsg:27700",
+        output_dir=tmpdir,
+        path_to_subgraph=sub_network["path"],
+        modes="bike",
+        increase_capacity=False,
+    )
+
+    for link_id, old_capacity in original_subnet_capacity.items():
```
mfitz, line 149: Again, it might help to extract the comparison into a helper function with a name that makes what we're asserting obvious but hides the details of how we perform that comparison.

tests/test_cli_replace_modal_subgraph.py
```diff
@@ -0,0 +1,195 @@
...
+        expected_new_link_id = sub_network["expected_link_id_mapping"][link_id]
+        assert output_network.has_link(
+            expected_new_link_id
+        ), f"Link {expected_new_link_id} is missing from the output network"
+        assert (
+            output_network.link(expected_new_link_id)["capacity"] == old_capacity
+        ), f"Link {expected_new_link_id} did not retain th original capacity value"
+
+
+def test_adds_nodes_only_once_with_multiple_subgraphs(tmpdir, network, sub_network):
```
mfitz, line 159: Technically, the same node could have been added multiple times, and this test would erroneously pass as long as the node count was as expected. If that is what we care about verifying, the test name should make that clear.

But if the test name accurately describes what we want to verify, the assertions should align with that intent. Right now, there is a disconnect between the intent of the test as suggested by the name, and the actual verification happening inside the test.

tests/test_cli_replace_modal_subgraph.py
```diff
@@ -0,0 +1,195 @@
...
+    output_network = cli._replace_modal_subgraph(
+        path_to_network=network["path"],
+        projection="epsg:27700",
+        output_dir=tmpdir,
+        path_to_subgraph=sub_network["path"],
+        modes="bike,bus",
+        increase_capacity=False,
+    )
+
+    assert len(list(output_network.links())) == original_net_no_links + (
```
mfitz, line 190: I can't understand from this test why we expect twice the number of links we see in the sub-network. Is there a way to make that clearer?

mfitz (changes requested):
I've made a few inline comments and asked a few questions that represent fairly trivial things.

I've also described a cleaner separation of concerns, where the CLI code would only be doing CLI things. This is a bigger change and does not need to happen in this PR, but it's something to bear in mind. Right now, the CLI and the - for want of a better phrase - "domain API" are smudged together when it would make more sense to decouple them.

